PR #6, opened 2018-04-14
Add no-transform and usqp.

twifkak:
Add Cache-Control: no-transform to the signed-exchange response, to
discourage intermediaries from modifying it.

Add usqp to the fetch request, to tell the AMP Cache to perform web
package transforms (in particular, additional headers).

Add a note that URL fragments are disallowed, per
https://github.com/WICG/webpackage/pull/170.

cc @nyaxt


nyaxt:
lgtm

Issue #22, opened 2018-06-20
Serve null validity map at /amppkg/validity


twifkak:
This is necessary for the same-origin policy imposed in https://crrev.com/c/1075833.

This is a sub-bug of #15.

PR #32, opened 2018-07-17
Serve a null validity map at /amppkg/validity

honeybadgerdontcare:
Addresses #22	

Also adds new validity url to the packager instead of the AMP CDN.
`validityUrl="http://{domain{:port}}/amppkg/validity"`

### DIFF
--- a/packager.go
+++ b/packager.go
@@ -218,8 +218,10 @@ func NewPackager(cert *x509.Certificate, key crypto.PrivateKey, packagerBase str
 	if !acceptablePackagerSchemes[baseURL.Scheme] {
 		return nil, errors.Errorf("PackagerBase %q must be over http or https.", packagerBase)
 	}
-	validityURL := baseURL
-	validityURL.Path = ValidityMapURL
+	validityURL, err := url.Parse(path.Join(packagerBase, ValidityMapURL))
+	if err != nil {
+		return nil, errors.Wrapf(err, "parsing PackagerBase %q with ValidityMapURL %q", packagerBase, ValidityMapURL)
+	}
 	client := http.Client{
 		// TODO(twifkak): Load-test and see if default transport settings are okay.
 		Timeout: 60 * time.Second,
--- a/packager.go
+++ b/packager.go
@@ -218,10 +218,8 @@ func NewPackager(cert *x509.Certificate, key crypto.PrivateKey, packagerBase str
 	if !acceptablePackagerSchemes[baseURL.Scheme] {
 		return nil, errors.Errorf("PackagerBase %q must be over http or https.", packagerBase)
 	}
-	validityURL, err := url.Parse("https://cdn.ampproject.org/null-validity")
-	if err != nil {
-		return nil, errors.Wrap(err, "parsing null-validity URL")
-	}
+	validityURL := baseURL
+	validityURL.Path = ValidityMapURL
 	client := http.Client{
 		// TODO(twifkak): Load-test and see if default transport settings are okay.
 		Timeout: 60 * time.Second,
--- a/packager.go
+++ b/packager.go
@@ -218,7 +218,8 @@ func NewPackager(cert *x509.Certificate, key crypto.PrivateKey, packagerBase str
 	if !acceptablePackagerSchemes[baseURL.Scheme] {
 		return nil, errors.Errorf("PackagerBase %q must be over http or https.", packagerBase)
 	}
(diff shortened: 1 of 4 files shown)


util.go
```diff
@@ -31,3 +31,6 @@ func CertName(cert *x509.Certificate) string {
 	sum := sha256.Sum256(cert.Raw)
 	return base64.RawURLEncoding.EncodeToString(sum[:])
 }
+
+// ValidityMapURL
```
twifkak: I take it this is required by the linter? (No action required, just venting.)

packager.go
```diff
@@ -218,10 +218,8 @@ func NewPackager(cert *x509.Certificate, key crypto.PrivateKey, packagerBase str
 	if !acceptablePackagerSchemes[baseURL.Scheme] {
 		return nil, errors.Errorf("PackagerBase %q must be over http or https.", packagerBase)
 	}
-	validityURL, err := url.Parse("https://cdn.ampproject.org/null-validity")
-	if err != nil {
-		return nil, errors.Wrap(err, "parsing null-validity URL")
-	}
+	validityURL := baseURL
+	validityURL.Path = ValidityMapURL
```
twifkak: baseURL may contain path components, e.g. `https://example.com/foobar/` -- setting Path will overwrite that. Please append instead, so it looks like `https://example.com/foobar/amppkg/validity`. I use ResolveReference in genCertUrl to do this, but there may be a better way (`path.Join` or something?).

validitymap.go
```diff
@@ -0,0 +1,41 @@
...
+func NewValidityMap() (*ValidityMap, error) {
+	this := new(ValidityMap)
+	this.validityMap = []byte("\xA0")
+	return this, nil
+}
+
+func (this ValidityMap) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	println("path", req.URL.Path)
+	if req.URL.Path == ValidityMapURL {
+		resp.Header().Set("Accept-Ranges", "bytes")
```
twifkak: You can use https://golang.org/pkg/net/http/#ServeContent which should "do the right thing" w.r.t. range requests. (Though you'll still need to specify content-type and cache-control separately.)

validitymap.go
```diff
@@ -0,0 +1,41 @@
...
+		resp.Header().Set("Content-Type", "text/html")
```
twifkak: Should be `application/cbor`.

validitymap.go
```diff
@@ -0,0 +1,41 @@
...
+	"net/http"
+)
+
+type ValidityMap struct {
+	validityMap []byte
+}
+
+func NewValidityMap() (*ValidityMap, error) {
+	this := new(ValidityMap)
+	this.validityMap = []byte("\xA0")
```
twifkak: Leave a comment that this is the empty map, and link to https://tools.ietf.org/html/draft-yasskin-httpbis-origin-signed-exchanges-impl-00#section-3.6.

util.go
honeybadgerdontcare: Updated since we care about the trailing slash in packagebase

packager.go
honeybadgerdontcare: Done

validitymap.go
honeybadgerdontcare: Done

validitymap.go
honeybadgerdontcare: Done

validitymap.go
honeybadgerdontcare: Done

packager.go
```diff
@@ -218,9 +218,9 @@ func NewPackager(cert *x509.Certificate, key crypto.PrivateKey, packagerBase str
 	if !acceptablePackagerSchemes[baseURL.Scheme] {
 		return nil, errors.Errorf("PackagerBase %q must be over http or https.", packagerBase)
 	}
-	validityURL, err := url.Parse("https://cdn.ampproject.org/null-validity")
+	validityURL, err := url.Parse(path.Join(packagerBase, ValidityMapURL))
```
twifkak: I think this has a bug: https://play.golang.org/p/UdOl82Y1BCL breaks the double-slash.

honeybadgerdontcare: Ah, yes, `path.Join` changes it to `http:/localhost/amppkg/validity` and `url.Parse` changes that to `http:///localhost/amppkg/validity`. Since `packagerBase` is guaranteed to have a trailing slash in `config.go`, using `+` for `packagerBase and `ValidityMapURL`.

PR #33, opened 2018-07-17
Do not cache error responses.

### LINKED ISSUE #20: Error messages should be non-cacheable
[HTTPError.LogAndRespond](https://github.com/ampproject/amppackager/blob/624a7f1edbea12b9ddb3ca739465ecca05dccb8b/errors.go#L40) calls [http.Error](https://golang.org/pkg/net/http/#Error), which [does not set](https://golang.org/src/net/http/server.go#L1956) `Cache-Control: no-store`. Update it to do so.

alin04:
Fixes #20

### DIFF
--- a/packager_test.go
+++ b/packager_test.go
@@ -96,7 +96,7 @@ func TestErrorNoCache(t *testing.T) {
 	urlSets := []URLSet{URLSet{
 		Fetch: &URLPattern{[]string{"http"}, "", "example.com", stringPtr("/amp/.*"), []string{}, stringPtr(""), false, boolPtr(true)},
 	}}
-	// Missign sign param generates an error.
+	// Missing sign param generates an error.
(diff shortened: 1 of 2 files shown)


packager_test.go
```diff
@@ -92,6 +92,18 @@ func TestNoFetchParam(t *testing.T) {
 	assert.Equal(t, "https://example.com/amp/secret-life-of-pine-trees.html", exchange.RequestUri.String())
 }
 
+func TestErrorNoCache(t *testing.T) {
+	urlSets := []URLSet{URLSet{
+		Fetch: &URLPattern{[]string{"http"}, "", "example.com", stringPtr("/amp/.*"), []string{}, stringPtr(""), false, boolPtr(true)},
+	}}
+	// Missign sign param generates an error.
```
twifkak: s/Missign/Missing

alin04: Done.

PR #34, opened 2018-07-17
pass conditional headers and pass back 304s

### LINKED ISSUE #19: Forward conditional request headers to the fetch URL
The packager should handle conditional requests.

1. The fetch issued by fetchURL() in packager.go should pass along any of the [conditional headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests#Conditional_headers) it received in ServeHTTP().

2. If the fetchResp is a 304 not modified, then ServeHTTP() should respond in kind, and not try to package up that 304 in an `application/signed-exchange`.

See also http://b/110430571.

honeybadgerdontcare:
Fixes #19 

There are no tests and probably should be but not sure how to test this with the current test framework that exists.


packager.go
```diff
@@ -43,6 +43,16 @@ const userAgent = "Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MMB29P) " +
 	"AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2272.96 Mobile " +
 	"Safari/537.36 (compatible; amppackager/0.0.0; +https://github.com/ampproject/amppackager)"
 
+// Conditional request headers that ServeHTTP may receive and need to be sent with fetchURL.
```
twifkak: This change seems significant enough to warrant a unittest. If you're ambitious, try adding one to packager_test.go. This will require changing the mock CDN that's set up in TestMain. (TestMain is a setup function that gets called once before all tests are run. Go doesn't have a traditional per-test "setup" method -- just include that code in the test itself.) If you're not ambitious, file a bug to implement the test. :)

packager.go
```diff
@@ -291,11 +308,23 @@ func (this Packager) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// Set conditional request header values
+	for header := range conditionalRequestHeaders {
```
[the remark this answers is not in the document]
honeybadgerdontcare: Done

packager.go
```diff
@@ -291,11 +308,23 @@ func (this Packager) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+
 	fetchReq, fetchResp, httpErr := this.fetchURL(fetchURL)
 	if httpErr != nil {
 		httpErr.LogAndRespond(resp)
 		return
 	}
+
+	// If fetchURL returns a 304, then also return a 304.
+	if fetchResp.StatusCode == 304 {
+		resp.WriteHeader(http.StatusNotModified)
```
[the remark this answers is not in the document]
honeybadgerdontcare: Done

packager.go
```diff
@@ -59,6 +69,19 @@ var statefulResponseHeaders = map[string]bool{
 	"WWW-Authenticate":          true,
 }
 
+// The server generating a 304 response MUST generate any of the
+// following header fields that would ahve been sent in a 200 (OK) response
+// to the same request.
+// https://tools.ietf.org/html/rfc7232#section-4.1
+statusNotModifiedHeaders = map[string]bool{
```
twifkak: Missing "var " in front.

packager.go
```diff
@@ -230,7 +253,7 @@ func NewPackager(cert *x509.Certificate, key crypto.PrivateKey, packagerBase str
 	return &Packager{cert, key, validityURL, &client, baseURL, urlSets}, nil
 }
 
-func (this Packager) fetchURL(orig *url.URL) (*http.Request, *http.Response, *HTTPError) {
+func (this Packager) fetchURL(orig *url.URL, serveHTTPReq *http.Request) (*http.Request, *http.Response, *HTTPError) {
```
twifkak: s/*http.Request/http.Header/

packager.go
```diff
@@ -291,11 +318,23 @@ func (this Packager) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
 		httpErr.LogAndRespond(resp)
 		return
 	}
+
+	// If fetchURL returns a 304, then also return a 304 with appropriate headers.
+	if fetchResp.StatusCode == 304 {
+		resp.WriteHeader(http.StatusNotModified)
+		for header := range statusNotModifiedHeaders {
+			if fetchResp.Header.Get(header) != "" {
+				resp.WriteHeader(fetchResp.Header.Get(header))
```
twifkak: This should be:

				resp.Header().Set(header, fetchResp.Header.Get(header))

packager.go
```diff
@@ -252,6 +275,10 @@ func (this Packager) fetchURL(orig *url.URL) (*http.Request, *http.Response, *HT
 		return nil, nil, NewHTTPError(http.StatusInternalServerError, "Error building request: ", err)
 	}
 	req.Header.Set("User-Agent", userAgent)
+	// Set conditional headers that were included in ServeHTTP's Request.
+	for header := range conditionalRequestHeaders {
+		req.Header.Set(header, serveHTTPReq.Get(header))
```
twifkak: Wrap this in `if serveHTTPReq.Get(header) != ""`; otherwise it sends empty headers for each of these.

honeybadgerdontcare: Ah, this one was in that newer commit. Thank you for the thoroughness.

packager.go
honeybadgerdontcare: git is acting up again so I want to be done with this branch. I've filed a bug #45 to address this feedback.

honeybadgerdontcare:
PTaL, `go test` fails with just

```
# github.com/ampproject/amppackager
./errors.go:31: missing ... in args forwarded to print-like function
FAIL	github.com/ampproject/amppackager [build failed]
```

packager.go
```diff
@@ -291,11 +320,23 @@ func (this Packager) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
-	fetchReq, fetchResp, httpErr := this.fetchURL(fetchURL)
+	fetchReq, fetchResp, httpErr := this.fetchURL(fetchURL, req.Header)
 	if httpErr != nil {
 		httpErr.LogAndRespond(resp)
 		return
 	}
+
+	// If fetchURL returns a 304, then also return a 304 with appropriate headers.
+	if fetchResp.StatusCode == 304 {
+		resp.WriteHeader(http.StatusNotModified)
```
twifkak: If my read of https://golang.org/pkg/net/http/#ResponseWriter is correct, then any headers you set after a call to `WriteHeader()` are ignored. I think you need to move this line after the for loop?

Also, I'm not positive, but I think you need to call `resp.Write([]byte{""})` in order to close the connection.

PR #41, opened 2018-08-13
Update README with local transformer section

alin04:


cmd/transform/transform.go
```diff
@@ -36,21 +34,19 @@ func main() {
...
 	}
 
 	flag.Parse()
-	var transformers []string
-	if *transformersFlag != "" {
-		transformers = strings.Split(*transformersFlag, ",")
-		for i := range transformers {
-			transformers[i] = strings.TrimSpace(transformers[i])
-		}
+	transformers := []string{
```
twifkak: (Eventually we'll probably want this somewhere that library-users can get it, too.)

PR #66, opened 2018-08-24
Add test that shows StartCron() will die if cache cannot be initialized.

alin04:


rtv_test.go
```diff
@@ -165,3 +165,27 @@ func TestRTVPollRollback(t *testing.T) {
...
+	origDie := die
+	defer func() { die = origDie }()
+	var errors string
+	die = func(format string, args ...interface{}) {
+		errors = fmt.Sprintf(format, args)
+	}
+	assert.Equal(t, "", RTVCache.RTV)
+	assert.Equal(t, "", RTVCache.CSS)
+	StartCron()
+	if errors == "" {
```
twifkak: Optionally, use assert.NotEmpty() for this.

Issue #14, opened 2018-05-18
Add OCSP stapling


twifkak:
http://crrev.com/c/1060933 adds a requirement that certUrls be OCSP-stapled, as part of the new v=b1 spec. The packager should have a flag for outputting b1 format, and doing the necessary stapling things. A set of requirements to start with is https://gist.github.com/sleevi/5efe9ef98961ecfb4da8, and a good implementation to look at is mholt/caddy/caddytls.

twifkak:
This is a child of #15.

PR #71, opened 2018-08-31
Fetch, cache, and attach OCSP responses.

twifkak:
This implements almost all of the requirements set by
https://gist.github.com/sleevi/5efe9ef98961ecfb4da8.

This uses a new Updateable interface, so
another transactional storage mechanism could be plugged in without any
architectural changes.

The code is, well, not pretty. It could use some cleanliness.

Addresses #14.


certcache.go
```diff
@@ -56,3 +108,245 @@ func (this *CertCache) ServeHTTP(resp http.ResponseWriter, req *http.Request, pa
...
+	}
+	if len(ocsp) == 0 {
+		return errors.New("Missing OCSP response.")
+	}
+	if !this.isHealthy(ocsp) {
+		return errors.New("OCSP failed health check.")
+	}
+	// Convert read lock into write lock before updating ocsp and ocspUpdateAfter.
+	this.ocspMuMu.Lock()
+	this.ocspMu.RUnlock()
```
alin04: In the above failure scenarios, is the read lock unlocked?

certcache.go
```diff
@@ -56,3 +108,245 @@ func (this *CertCache) ServeHTTP(resp http.ResponseWriter, req *http.Request, pa
...
+	}
+	this.ocspMu.Unlock()
+	return nil
+}
+
+// Checks for OCSP updates every hour. Never terminates.
+func (this *CertCache) maintainOCSP(stop chan struct{}) {
+	// Only make one request per ocspCheckInterval, to minimize the impact
+	// on OCSP servers that are buckling under load (sleevi #5).
+	ticker := time.NewTicker(ocspCheckInterval)
```
alin04: Could leverage the cron that rtv.go uses. Refactor to pull out the cron var so this new job can be added to it...

certcache.go
```diff
@@ -16,24 +16,74 @@ package amppackager
...
+
+// 1MB is the maximum used by github.com/xenolf/lego/acmev2 in GetOCSPForCert.
+// Alternatively, here's a random documented example of a 20K limit:
+// https://www.ibm.com/support/knowledgecenter/en/SSPREK_9.0.0/com.ibm.isam.doc/wrp_stza_ref/reference/ref_ocsp_max_size.html
+const maxOCSPResponseBytes = 1024 * 1024
+
+// How often to check if OCSP stapling needs updating.
+const ocspCheckInterval = 1 * time.Hour
+
+var fakeOCSPServer string
```
alin04: "fake" in code? code smell?

storage.go
```diff
@@ -0,0 +1,81 @@
...
+	path string
+}
+
+func (this LocalFile) Read(ctx context.Context, isExpired func([]byte) bool, update func([]byte) []byte) ([]byte, error) {
+	lock := flock.NewFlock(this.path)
+	locked, err := lock.TryRLock()
+	if !locked {
+		return nil, errors.Errorf("unable to obtain shared lock for %s", this.path)
+	}
+	if err != nil {
```
alin04: Check for err first?

certcache.go
twifkak: Good catch. I made this error elsewhere, too (including storage.go). Fixed (as soon as I push).

certcache.go
twifkak: I think this has a few benefits over the `robfig/cron` library:
 - One fewer dependency.
 - The syntax is checked at compile-time, rather than "@every 1h" which happens at run-time. That said: 1) The actual logic isn't checked, 2) If we need more complicated scheduling (e.g. "every tuesday") then cron is probably better, and 3) It looks like cron has a `ConstantDelaySchedule` type that address my complaint. So I'm on the fence on this one.
 - With this format, main.go and unit tests can start/stop only the schedules they care about. With cron, you can basically only add a bunch of funcs during initialization. If you want to change the funcs that run, you need to replace the whole Cron to clear it (AFAICT). But maybe that's good enough for tests.

WDYT?

certcache.go
twifkak: Totally a code smell. I added a comment and am updating these to be struct fields (work-in-progress) rather than globals, but I don't see a way around them.

OCSP server is normally extracted from the certificate, which is static [1], but we need to point it at a fake server, which gets a dynamically-assigned port from the httptest library.

OCSP expiry is normally computed from the HTTP response headers, by the 3rd party cachecontrol library. That library uses `time.Now` internally, so we need to stub out our use of it for test.

[1] I guess we could mint certs in test setup, but eww.

storage.go
twifkak: Done (as soon as I push).

certcache.go
```diff
@@ -56,3 +108,245 @@ func (this *CertCache) ServeHTTP(resp http.ResponseWriter, req *http.Request, pa
...
+		return false
+	}
+	issuer := this.findIssuer()
+	if issuer == nil {
+		log.Println("Cannot find issuer certificate in CertFile.")
+		return false
+	}
+	resp, err := ocsp.ParseResponseForCert(ocspResp, this.certs[0], issuer)
+	if err != nil {
+		log.Println("Error parsing OCSP response:", err)
```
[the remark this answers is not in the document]
twifkak: Actually, no: https://golang.org/pkg/fmt/#Println

Though, for funzies, Print and Printf in that package both handle spaces slightly differently: https://golang.org/pkg/fmt/#Print

certcache.go
```diff
@@ -56,3 +108,245 @@ func (this *CertCache) ServeHTTP(resp http.ResponseWriter, req *http.Request, pa
...
+		return errors.New("Missing OCSP response.")
+	}
+	if !this.isHealthy(ocsp) {
+		return errors.New("OCSP failed health check.")
+	}
+	// Convert read lock into write lock before updating ocsp and ocspUpdateAfter.
+	this.ocspMuMu.Lock()
+	this.ocspMu.RUnlock()
+	this.ocspMu.Lock()
+	this.ocspMuMu.Unlock()
```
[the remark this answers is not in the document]
twifkak: You are totally correct. I deleted ocspMuMu so now it's simply "acquire read lock, release, acquire write lock, release", and re-read the file inside the write-lock per your suggestion. (In the next push.)

certcache.go
alin04: As long as we're consistent and use one way. So either hange this to cron or change rtv to use NewTicker (i.e. add TODO for whatever is decided)

I think we can get around the cron initialization/testing through clever coding.

certcache.go
```diff
@@ -16,24 +16,74 @@ package amppackager
...
 }
 
-func NewCertCache(cert *x509.Certificate) (*CertCache, error) {
+func NewCertCache(certs []*x509.Certificate, ocspCache string, stop chan struct{}) (*CertCache, error) {
 	this := new(CertCache)
-	this.certName = CertName(cert)
-	this.cert = cert
+	this.certName = CertName(certs[0])
+	this.certs = certs
+	// sleevi #1, sleevi #6:
```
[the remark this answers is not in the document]
twifkak: Done (as soon as I push). Though I'd recommend referencing the sleevi doc in full while reviewing this code, as my snippets don't contain all the nuance (especially for `#2`).

certcache.go
twifkak: OK, I added a TODO to change rtv, as long as that's OK by you. Partly motivated by trying to do things "the Go way", and partly by not knowing what robfig/cron does (e.g. I started perusing and noticed that it eats panics from the jobs it runs).

twifkak:
PTAL. I rebased, but the only new stuff is in the last commit.

certcache.go
```diff
@@ -16,25 +16,102 @@ package amppackager
...
+	// an hour at most.
+	ocspMu          sync.RWMutex
+	ocsp            []byte
+	ocspUpdateAfter time.Time
+	// TODO(twifkak): Implement a registry of Updateable instances which can be configured in the toml.
+	ocspFile Updateable
+	client   http.Client
+
+	// Fakes exposed for testing. Should be zero in production.
+	fakeOCSPServer string
```
alin04: How about creating a func assigned to a var that can be overridden in your test?

in real code:
var getOCSPServerOrDefault = func(certs []x509.Certificate, orig []byte) {
    if len(this.certs[0].OCSPServer) < 1 {
        log.Println("Cert missing OCSPServer.")
￼ 	return orig
    } else {
￼ 	// This is a URI, per https://tools.ietf.org/html/rfc5280#section-4.2.2.1.
￼  	return this.certs[0].OCSPServer[0]
    }
}

In test:
var getOCSPServer = func(_ []*x509.Certificate, _ []byte) {
  return fake
}

twifkak: Done.

twifkak:
PTAL. I got rid of the `fake` vars, and extracted the `ocspMu` stuff outside of the `CertCache` struct. (There's still a mutex for `ocspUpdateAfter`, as I'm trying to keep `Updateable` as a pure `[]byte` thing. But I might give up and make it `interface{}` instead.)

certcache.go
```diff
@@ -56,3 +146,241 @@ func (this *CertCache) ServeHTTP(resp http.ResponseWriter, req *http.Request, pa
...
+	return ocsp, ocspUpdateAfter, nil
+}
+
+// Checks for OCSP updates every hour. Never terminates.
+func (this *CertCache) maintainOCSP(stop chan struct{}) {
+	// Only make one request per ocspCheckInterval, to minimize the impact
+	// on OCSP servers that are buckling under load, per sleevi requirement:
+	// 5. As with any system doing background requests on a remote server,
+	//    don't be a jerk and hammer the server when things are bad...
+	//    sometimes servers and networks have issues. When a[n OCSP client]
```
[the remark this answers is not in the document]
twifkak: No, paraphrasing the original quote.

storage.go
```diff
@@ -0,0 +1,142 @@
...
+		// multiple calls to update() in parallel.
+		contents, err := ioutil.ReadFile(this.path)
+		if err != nil {
+			return nil, errors.Wrapf(err, "rereading %s", this.path)
+		}
+		if !isExpired(contents) {
+			return contents, nil
+		}
+		contents = update(contents)
+		// TODO(twifkak): Should I write to a tempfile in the same dir and move into place, instead?
```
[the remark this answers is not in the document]
twifkak: Fair enough. I'll not worry about it for now, but leave it until if/when I decide to spend some time researching.

PR #77, opened 2018-09-04
Pass rtv to transformation lib.

alin04:


packager.go
```diff
@@ -91,7 +91,8 @@ const miRecordSize = 4096
 
 // Overrideable for testing.
 var getTransformerRequest = func(s, u string) *rpb.Request {
-	return &rpb.Request{Html: string(s), DocumentUrl: u}
+	r := &rpb.Request{Html: string(s), DocumentUrl: u, Rtv: RTVCache.RTV, Css: RTVCache.CSS}
```
twifkak: I'd suggest wrapping reads and writes to RTVCache in a `sync.RWMutex`, per the advice here: https://golang.org/ref/mem#tmp_1

It might be okay, per the "Happens Before" section that follows, but I'm not positive I'm reading that right.

(On a side note: Perhaps the use of mutexes is a code smell in Go. I think the [idiomatic way](https://golang.org/doc/faq#What_operations_are_atomic_What_about_mutexes) involves using channels to communicate state changes. But that's a change for another day.)

alin04:
PTAL. I changed quite a few things.

packager.go
alin04: Ah, I only did a Mutex. Can switch to RWMutex in future CL.

rtv.go
```diff
@@ -5,89 +5,111 @@ import (
...
+	r.lk.Lock()
+	defer r.lk.Unlock()
+	return r.d.css
+}
+
+// poll attempts to re-populate the RTVCache, returning an error if there
+// were any problems.
+func (r *RTVCache) poll() error {
+	// Make a copy for atomicity.
+	d := *r.d
```
twifkak: I believe this read should be behind a lock as well. (Maybe make a getD() func.)

alin04: Done

Issue #28, opened 2018-06-21
Reorganize directories to match standard go layout


twifkak:
Follow the [Standard Go Project Layout](https://github.com/golang-standards/project-layout). That entails:

- moving the `.go` files from the root into `internal/pkg/`
- moving the fuzz tests into `test/`
- moving the example toml into `configs/`

and maybe some other stuff.

alin04:
Propose segregating packager from transformer code, e.g.
packager/
transformer/
Or should they be verbs?
package/
transform/

honeybadgerdontcare:
I'd go with what we plan on calling it in documentation, which most likely will be nouns not verbs.

twifkak:
Nouns SGTM. I'm slightly in favor of dropping 'pkg' and just having the code be at `transformer/**/*.go` or `transformer/**/internal/*.go`. https://github.com/golang-standards/project-layout/tree/master/pkg admits it's not universally accepted anyway. None of the third party libs we have in `vendor/` do this. But that's not a strongly held opinion; thoughts?

Once `pkg` is renamed to `transformer`, it may be worth renaming `transformer/transformer` to `transformer/transformers`, and moving `transformer/transform/transform.go` up a level to `transformer/transformer.go`. So people looking at this directory for the first time may start in the right place. :)

twifkak:
Still needs #88 (or equivalent).

alin04:
Hmm, I'm still on the fence about noun vs verb. Both are applicable depending on the circumstance. E.g. go's html package has both: node.go, token.go (verbs). render.go, parse.go (verbs)

So, perhaps
  transform/transformers/...
  transform/transform.go 

Bikeshedding...  You can ignore me. :)

PR #87, opened 2018-09-12
Update dependent code from renames in #86.

twifkak:
Only pkg/amp_validator remains, and that will be moved to
ampproject/amphtml#18006.


twifkak:
Addresses #28.

PR #88, opened 2018-09-12
Depend on validator.pb.go from amphtml project.

twifkak:
This removes the remaining file in pkg/. Also, add `prune.non-go = true`
to Gopkg.toml, to eliminate a bunch of unnecessary files in vendor/.

Addresses #28.


twifkak:
PTAL - This is what's leftover from #92, and no longer covers `transformer/`.

twifkak:
Fixes #28.

twifkak:
That's in this commit; it just shows up as a rename. See https://github.com/ampproject/amppackager/pull/88/files#diff-94ee56b283d614195a3f7fd45652a572 or https://github.com/ampproject/amppackager/pull/88.patch.

PR #90, opened 2018-09-13
Remove TestEOFNewline.

twifkak:
This was broken in 463a9de3 (#75).


alin04:
oh no. google should be SoT

twifkak:
oh no indeed. i'll revert this and do it again in google.

PR #92, opened 2018-09-13
Sync from Google

twifkak:
This is an automatically generated pull request.


twifkak:
This change corresponds to cl/212717811 and cl/212753076, from Google.

PR #103, opened 2018-09-15
Fallback to unsigned if AMP-Cache-Transform wrong.

### LINKED ISSUE #84: Fallback to unsigned if AMP-Cache-Transform is wrong
This should parse https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-transform.md from the request and ensure target is `any` or `google`, and no params are specified.

twifkak:
For now, this means that it should contain an unparameterised `any` or
`google`.

Fixes #84.

### DIFF
--- a/packager/amp_cache_transform/amp_cache_transform.go
+++ b/packager/amp_cache_transform/amp_cache_transform.go
@@ -1,0 +1,139 @@
+// Parses and interprets the AMP-Cache-Transform header specified at
+// https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-transform.md.
+
+package amp_cache_transform
+
+import (
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.3
+type parameterisedIdentifier struct {
+	id     string
+	// In the future, this will include params.
+}
+
+// https://tools.ietf.org/html/rfc7230#appendix-B (OWS = "optional whitespace")
+func discardOWS(reader *strings.Reader) error {
+	for reader.Len() > 0 {
+		c, err := reader.ReadByte()
+		if err != nil {
+			return errors.Wrap(err, "reading OWS")
+		}
+		if c != ' ' && c != '\t' {
+			// Return to the position before the non-whitespace char.
+			reader.Seek(-1, io.SeekCurrent)
+			return nil
+		}
+	}
+	return nil
+}
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.3
+func parseParameterisedList(reader *strings.Reader) ([]parameterisedIdentifier, error) {
+	items := []parameterisedIdentifier{}
+	for reader.Len() > 0 {
+		item, err := parseParameterisedIdentifier(reader)
+		if err != nil {
+			return nil, errors.Wrap(err, "parsing parameterised identifier")
+		}
+		items = append(items, *item)
+		// Steps 3 and 4 are swapped per https://github.com/httpwg/http-extensions/pull/667.
+		if reader.Len() == 0 {
+			return items, nil
+		}
+		if err := discardOWS(reader); err != nil {
+			return nil, errors.Wrap(err, "discarding OWS")
+		}
+		comma, err := reader.ReadByte()
+		if err != nil {
+			return nil, errors.Wrap(err, "reading comma")
+		}
+		if comma != ',' {
+			return nil, errors.New("expected comma")
+		}
+		if err := discardOWS(reader); err != nil {
+			return nil, errors.Wrap(err, "discarding OWS")
+		}
+		if reader.Len() == 0 {
+			return nil, errors.New("expected another param-id")
+		}
+	}
+	return nil, errors.New("reached unexpected end of parseParameterisedList")
+}
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.4
+func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentifier, error) {
+	primaryID, err := parseIdentifier(reader)
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing primary identifier")
+	}
+	// NOTE: The initial version of AMP-Cache-Transform does not support
+	// any parameters, so we don't bother to implement a parser for them.
+	// No point in distinguishing syntactic failure from semantic failure.
+	// Once we add parameters, we'll need to implement parsers for the
+	// subset item that they use. See Items here:
+	// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.4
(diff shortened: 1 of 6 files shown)


packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -0,0 +1,139 @@
...
+			return nil, errors.New("expected comma")
+		}
+		if err := discardOWS(reader); err != nil {
+			return nil, errors.Wrap(err, "discarding OWS")
+		}
+		if reader.Len() == 0 {
+			return nil, errors.New("expected another param-id")
+		}
+	}
+	return nil, errors.New("reached unexpected end of parseParameterisedList")
```
Gregable: nit. The only way you can exit the for loop without returning is if the input reader is empty to begin with, right?

There are no break statements, and the end of the for loop body return exits if the for loop condition would fail.

It seems that this error message could be more specific ("empty parameterized list") or you could eliminate the last "len() = 0" check at the end of the for body and make this generic.

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -0,0 +1,139 @@
...
+	// any parameters, so we don't bother to implement a parser for them.
+	// No point in distinguishing syntactic failure from semantic failure.
+	// Once we add parameters, we'll need to implement parsers for the
+	// subset item that they use. See Items here:
+	// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.4
+	return &parameterisedIdentifier{primaryID}, nil
+}
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.8
+func isIdentifier(c byte) bool {
```
Gregable: naming. isIdentifierByte? isIdentifierChar?

Oddly named. It neither handles the first byte of an identifier (it's too lax, as that can only be lcalpha) nor does it verify the whole identifier.

Gregable: also nit, but maybe put this below the isdigit and islcalpha f'ns, since it depends on them, and places this method closer to it's only dependent.

packager/amp_cache_transform/amp_cache_transform_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestShouldSendSXG(t *testing.T) {
+	assert.True(t, ShouldSendSXG("any"))
+	assert.True(t, ShouldSendSXG("google"))
+	assert.True(t, ShouldSendSXG("google, any"))
+	assert.True(t, ShouldSendSXG("google, foobar"))
+	assert.True(t, ShouldSendSXG("foobar, any"))
```
Gregable: "google,foobar"

Verify that space isn't significant. Also try with "google,\tfoobar" "google ,foobar"

assert.False(t, ShouldSendSXG("googleany"))

Verify that the code isn't just performing substring matching.

Verify some of the error checking for identifiers:

False: "google,123"
False: "ABC,google"

twifkak: Done.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Done and done.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Done. I tried to copy the spec'd algorithm as closely as possible, and for the life of me, couldn't figure out what this error was even doing! Case of the Fridays. I'd mistakenly thought it was an "unreachable" kind of thing.

twifkak:
I'll wait for @alin04 to approve.

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -0,0 +1,139 @@
...
+			// Return to the position before the non-whitespace char.
+			reader.Seek(-1, io.SeekCurrent)
+			return nil
+		}
+	}
+	return nil
+}
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.3
+func parseParameterisedList(reader *strings.Reader) ([]parameterisedIdentifier, error) {
```
alin04: Ah, the Queen's English now, eh? j/k. :)

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -0,0 +1,139 @@
...
+	// any parameters, so we don't bother to implement a parser for them.
+	// No point in distinguishing syntactic failure from semantic failure.
+	// Once we add parameters, we'll need to implement parsers for the
+	// subset item that they use. See Items here:
+	// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.4
+	return &parameterisedIdentifier{primaryID}, nil
+}
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.8
+func isLCAlpha(c byte) bool {
```
alin04: unicode.IsLower() ? (maybe that includes more than standard ascii)

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -0,0 +1,139 @@
...
+	return c >= 'a' && c <= 'z'
+}
+
+// https://tools.ietf.org/html/rfc5234#appendix-B.1
+func isDigit(c byte) bool {
```
alin04: unicode.IsDigit()

packager/packager.go
```diff
@@ -338,20 +339,39 @@ func (this *Packager) ServeHTTP(resp http.ResponseWriter, req *http.Request, par
...
 	}()
 
-	if !this.shouldPackage() {
+	if !this.shouldPackage() || !amp_cache_transform.ShouldSendSXG(req.Header.Get("AMP-Cache-Transform")) {
 		proxy(resp, fetchResp)
 		return
 	}
 
 	switch fetchResp.StatusCode {
+	case 200:
```
alin04: Ah, I missed doing this in my last commit. Thanks for cleaning up.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Yeah, this is different. The impl in https://golang.org/src/unicode/letter.go?s=5508:5533#L182 depends on a generated `properties` var. The first difference is at https://golang.org/src/unicode/letter.go?s=5508:5533#L182, and there are a bunch of other extraneous pLl instances below that.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: This is technically correct, per https://golang.org/src/unicode/digit.go?s=233:258#L7 and https://golang.org/src/unicode/letter.go#L16. But it would require converting a byte into a rune, and also I prefer having the code right here so it's obviously equal to the spec definition linked in the comment.

packager/packager.go
twifkak: No prob!

PR #104, opened 2018-09-17
Override fetched response headers

### LINKED ISSUE #36: Override fetched response headers
At a minimum, it needs to specify:
 * `Content-Security-Policy`
 * `Content-Type: text/html`

twifkak:
Override Content-Security-Policy, Content-Type, and Link from the
fetched response, as these are the most security-relevant.

Fixes #36.

### DIFF
--- a/packager/packager.go
+++ b/packager/packager.go
@@ -26,6 +26,7 @@ import (
 	"path"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/WICG/webpackage/go/signedexchange"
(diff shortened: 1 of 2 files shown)


packager/packager.go
```diff
@@ -363,6 +374,11 @@ func (this *Packager) ServeHTTP(resp http.ResponseWriter, req *http.Request, par
 			fetchResp.Header.Del(header)
 		}
 
+		// charset=utf-8 would be redundant, as it is specified in the <meta> of a valid AMPHTML document:
+		fetchResp.Header.Set("Content-Type", "text/html")
+		fetchResp.Header.Set("Content-Security-Policy", contentSecurityPolicy)
+		fetchResp.Header.Del("Link")  // Ensure there are no privacy-violating Link:rel=preload headers.
```
Gregable: Maybe also strip `Location`. Shouldn't matter if we're serving a 200, but defense in depth.

Also `Set-Cookie`. Pubs may mess this up and automatically set cookies on all cookieless requests. We don't want to hand multiple users the same identifier.

twifkak: Meh, `Location` is apparently harmless on 200, and the browser will reject signed redirects per https://wicg.github.io/webpackage/loading.html#parsing-b1 so there's no chance it'll prefetch through them. It's useless but also harmless, and I lean towards less code in general. I could be swayed.

`Set-Cookie` is stripped already, see `statefulResponseHeaders`.

PR #107, opened 2018-09-17
Change strings.Builder to bytes.Buffer.

twifkak:
This allows the packager to run on Go 1.8, which is the current version
of Go on https://wiki.debian.org/DebianStretch, the current "stable"
version of Debian.

Ignore the "b2" commit; this will be rebased away.


twifkak:
Nevermind, this doesn't work anyway because `vendor/github.com/WICG/webpackage` has a dependency on Go 1.9 for `asn1.NullBytes`, and Go 1.10 is obtainable on https://backports.debian.org/.

PR #113, opened 2018-09-19
Remove PackagerBase.

### LINKED ISSUE #79: Improve documentation
Rewrite the documentation to be action-oriented -- e.g. start with running the packager, and then move backwards up the chain to adding a FE and all that.

twifkak:
It is confusing, and also wrong, now that the validity-url must be
same-origin with the signed URL. Instead, the behavior is thus:

1. validity-url is computed using the origin of the signed URL. It
   doesn't matter if this doesn't actually exist; the browser doesn't
   currently fetch it.
2. In development, the cert-url is localhost, so that you can test with
   a browser.
3. In production, the cert-url is computed using the origin of the
   signed URL. The README.md states that publishers must configure this
   when productionizing.

Fixes #79.

Ignore the first 4 commits; they will be rebased away.

### DIFF
--- a/amppkg.example.toml
+++ b/amppkg.example.toml
@@ -7,25 +7,6 @@
 # binding on the loopback interface.
 # LocalOnly = true
 
-# This is the URL prefix under which the packager is being served on the open
-# internet. This is used to generate cert-urls and validity-urls, by appending
-# "/amppkg/cert/blahblahblah" and "/amppkg/validity". This is
-#
-# For instance, if PackagerBase = "https://example.com/", then the frontend
-# server should reverse-proxy all URLs matching:
-#   https://example.com/amppkg/(.*)
-# to the packager.
-#
-# Alternatively, if PackagerBase = "https://example.com/packager/", then the
-# frontend server should reverse-proxy all URLs matching:
-#   https://example.com/packager/amppkg/(.*)
-# to the packager, removing the "/packager" prefix from the path.
-#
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -27,165 +27,138 @@ and signs AMP documents as requested by the AMP Cache.
...
+  3. Stop `amppkg` with Ctrl-C.
+  4. `go get -u github.com/ampproject/amppackager/cmd/amppkg_test_cache`.
+  5. `amppkg_test_cache`
+  6. Open Chrome and DevTools.
+  7. Visit `https://localhost:8000/`. Observe the prefetch of `/test.sxg`.
+  8. Click the link. Observe that the cached SXG is used.
+
+#### Productionizing
+
+For now, productionizing is a bit manual. We would love to see scripts or
```
alin04: Maybe omit this last sentence?

README.md
```diff
@@ -27,165 +27,138 @@ and signs AMP documents as requested by the AMP Cache.
...
+You may also want to:
+
+  1. Launch `amppkg` as a restricted user.
+  2. Save its stdout to a rotated log somewhere.
+
+Once you've done the above, you should be able to launch Chrome without any
+comamndline flags; just make sure chrome://flags/#enable-signed-http-exchange is
+enabled.
+
+#### Replicating
```
alin04: Redundancy?  I always think of replication in the context of data replication...

PR #116, opened 2018-09-21
Require CanSignHttpExchanges in production.

### LINKED ISSUE #110: Verify `CanSignHttpExchanges` is present
If `-development` is not true, then the packager should refuse to run or sign exchanges if given a non-SXG cert.

twifkak:
This prevents users from accidentally using their TLS cert pairs to sign
exchanges, as such SXGs would get rejected by the browser.

Fixes #110.

### DIFF
--- a/cmd/amppkg/main.go
+++ b/cmd/amppkg/main.go
@@ -19,7 +19,6 @@ package main
 
 import (
 	"bytes"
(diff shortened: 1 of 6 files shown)


twifkak:
OK, this PR is clean now (safe to review all changes).

Issue #62, opened 2018-08-24
Make it easier for new devs


twifkak:
- [x] Check in a self-signed cert/key pair and toml
 - [x] Update CONTRIBUTING.md with instructions on running a dev server, lint/tests, `dep ensure`, etc.
 - [x] Presubmit hook would be nice.

twifkak:
Travis hook is now working!

PR #120, opened 2018-09-22
Add minimal Travis config.

twifkak:
This configures Travis to run `go test` on every PR, as soon as the repo
is Activated in Travis.

Addresses #62.


twifkak:
Yes, the [default command](https://docs.travis-ci.com/user/languages/go/#default-build-script) of `go test -v ./...` is exactly what we want.

Issue #17, opened 2018-06-19
Improve test coverage


twifkak:

twifkak:
- [ ] `packager/util/config.go`
 - [ ] `packager/util/errors.go`
 - [ ] `packager/util/util.go`: `CertName` and `ParsePrivateKey`
 - [ ] `packager/validitymap/validitymap.go`
 - [ ] `packager/signer/signer.go`: `parseURL`, `fetchUrlMatches`, `parseUrlMatches`, `urlsMatch`, `parseURLs`, `validateFetch`

TODO: Add more from `signer.go`.

twifkak:
`signer.go`:
 - [ ] signature is correct (including cert-url, validity-url)
 - [ ] fetch/sign passed in post body
 - [ ] amp-cache-transform invalid
 - [ ] accept invalid
 - [ ] stateful header deleted if !errorOnStatefulHeaders

also, convert this to use testify/suite.

PR #136, opened 2018-10-04
Add more signer tests.

twifkak:
These test all the free-standing URL-validation functions as well as the
fetch-response validation function. Also changes validateFetch() to use
mime.ParseMediaType instead of a hacky string comparison.

Addresses #17.

### DIFF
--- a/packager/signer/validation.go
+++ b/packager/signer/validation.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pquerna/cachecontrol"
 )
 
+// Converts an URL string into an URL object with an unambiguous interpretation.
 func parseURL(rawURL string, name string) (*url.URL, *util.HTTPError) {
 	if rawURL == "" {
 		return nil, util.NewHTTPError(http.StatusBadRequest, name, " URL is unspecified")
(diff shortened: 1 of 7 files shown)


packager/util/errors_test.go
```diff
@@ -24,6 +24,10 @@ func (this *ErrorsSuite) TearDownTest() {
 	log.SetOutput(os.Stderr)
 }
 
+func (this *ErrorsSuite) TestError() {
+	this.Assert().Equal("Coffee grinder is broken", NewHTTPError(418, "Coffee grinder is broken").Error())
```
honeybadgerdontcare: But it's a teapot, not a grinder!

packager/signer/validation.go
```diff
@@ -0,0 +1,165 @@
...
+			return errors.Errorf("PathExcludeRE matches: %s", re)
+		}
+	}
+	if !regexpFullMatch(*pattern.QueryRE, url.RawQuery) {
+		return errors.New("QueryRE doesn't match")
+	}
+	return nil
+}
+
+func schemeMatches(actual string, expecteds []string) bool {
```
honeybadgerdontcare: Instead of `expecteds` perhaps `schemes` or `expected_schemes`, though I know Go prefers brevity with naming.

twifkak: Done.

twifkak:
Yeah, totally fair. Added some high-level comments for all the functions plus a few for function innards. If there's anything in particular you think needs clarifying, LMK.

PR #141, opened 2018-10-05
Update README.md

honeybadgerdontcare:
Chrome shows it's a signed exchange via the type column in DevTools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,8 @@ can obtain certificates for.
   2. Open DevTools. Check 'Preserve log'.
   3. Click the `click me` link.
   4. Watch the URL transmogrify! Verify it came from an SXG by switching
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,8 +72,7 @@ can obtain certificates for.
   2. Open DevTools. Check 'Preserve log'.
   3. Click the `click me` link.
   4. Watch the URL transmogrify! Verify it came from an SXG by switching
-     DevTools to the Network tab and looking in the Size column for "(from
-     signed-exchange)".
+     DevTools to the Network tab and looking in the Type column for `signed-exchange`.
```
twifkak: How about both? The "outer" response shows Type: signed-exchange, while the "inner response" two rows down should show Size: (from signed-exchange), which should be an indication that the SXG actually verified.

honeybadgerdontcare: Done

PR #144, opened 2018-10-05
Add nosniff headers.

### LINKED ISSUE #138: Add nosniff header
Add `X-Content-Type-Options: nosniff` to the SXG outer response headers.

twifkak:
Add `X-Content-Type-Options: nosniff` response header to SXG outer, SXG
inner, cert-url, and validity-url.

Fixes #138.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -288,6 +288,7 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 	//  2. This would require some logic for determining media type
 	//     equivalence (including parameters).
 	fetchResp.Header.Set("Content-Type", "text/html")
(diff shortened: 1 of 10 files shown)


packager/signer/signer.go
```diff
@@ -349,6 +350,7 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 	// should fetch an update (half-way between signature date & expires).
 	resp.Header().Set("Content-Type", "application/signed-exchange;v=b2")
```
alin04: I wonder if these magic values should be replaced by const? E.g. when applicable, easier to change v=b2 to v=b3 at the top of the file instead of hunting around.

Also, maybe easier to understand if resp and fetchResp are renamed? outer? inner? sxgResp?

Issue #140, opened 2018-10-05
Perform light AMP validation


twifkak:
Transforms may have unintended effects on invalid documents. There may be cases where the SXG is fetched and served by third parties without validation. Since full validation at serve time is too expensive, it should perform some light validation; at least to verify that the document intends to be AMP.

Gregable:
Knowing the transforms that run, I doubt any of the transforms would make an invalid amp document particularly dangerous. Hard to guarantee that's the case, but most of them just exist to make the document *less* likely to behave in non-amp manners, such as executing javascript.

An additional idea would be to add a transformer that strips a handful of known disallowed HTML, such as extra attention to `script` tags and stripping `on*` attributes. These will be no-ops on amp pages, and so should be cheap to run.

twifkak:
@Gregable's suggestion has been split into #148.

PR #152, opened 2018-10-09
Add Request_AMP to transformer request.

twifkak:
Tell the transformer to transform only if the document declares itself
AMP. This overrides the default that it will do so also for AMP4ADS and
AMP4EMAIL.

Also, fix the call to proxy() from serveSignedExchange() (in cases where
the transformer errors) to proxy the response body as well.

Addresses #140.

### DIFF
--- a/packager/signer/signer_test.go
+++ b/packager/signer/signer_test.go
@@ -136,7 +136,8 @@ func (this *SignerSuite) SetupTest() {
 	}
 	// Don't actually do any transforms. Only parse & print.
 	getTransformerRequest = func(r *rtv.RTVCache, s, u string) *rpb.Request {
-		return &rpb.Request{Html: string(s), DocumentUrl: u, Config: rpb.Request_NONE}
+		return &rpb.Request{Html: string(s), DocumentUrl: u, Config: rpb.Request_NONE,
+			AllowedFormats: []rpb.Request_HtmlFormat{rpb.Request_AMP}}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


packager/signer/signer_test.go
```diff
@@ -380,18 +381,56 @@ func (this *SignerSuite) TestProxyUnsignedNonCachable() {
 	this.Assert().Equal("text/html", resp.Header.Get("Content-Type"))
 }
 
+func (this *SignerSuite) TestProxyUnsignedIfNotAMP() {
+	urlSets := []util.URLSet{{
+		Sign: &util.URLPattern{[]string{"https"}, "", this.httpsHost(), stringPtr("/amp/.*"), []string{}, stringPtr(""), false, nil}}}
+	nonAMPBody := []byte("<html><body>They like to OPINE. Get it? (Is he fir real? Yew gotta be kidding me.)")
+	this.fakeHandler = func(w http.ResponseWriter, req *http.Request) {
```
Gregable: Naming: Can we use "writer" or something better than "w"? Optional as I know this code already does this a lot.

packager/signer/signer_test.go
```diff
@@ -380,18 +381,56 @@ func (this *SignerSuite) TestProxyUnsignedNonCachable() {
...
+	body, err := ioutil.ReadAll(resp.Body)
+	this.Require().NoError(err)
+	this.Assert().Equal(nonAMPBody, body, "incorrect body: %#v", resp)
+}
+
+func (this *SignerSuite) TestProxyUnsignedIfWrongAMP() {
+	urlSets := []util.URLSet{{
+		Sign: &util.URLPattern{[]string{"https"}, "", this.httpsHost(), stringPtr("/amp/.*"), []string{}, stringPtr(""), false, nil}}}
+	wrongAMPBody := []byte("<html amp4email><body>They like to OPINE. Get it? (Is he fir real? Yew gotta be kidding me.)")
+	this.fakeHandler = func(w http.ResponseWriter, req *http.Request) {
```
Gregable: ditto

packager/signer/signer_test.go
twifkak: OK, renamed to `resp` as that's pretty pervasive throughout the codebase.

PR #154, opened 2018-10-12
Sync from Google

twifkak:
This includes 4 automatically generated commits, and 1 manual commit to make Travis happy.


twifkak:
Unknown committer `Googler` is a Googler, and thus good for CLA.

PR #158, opened 2018-10-17
Add inner Link rel=preload header.

### LINKED ISSUE #37: Add `Link: rel=preload`s
Add relevant preloads as determined based on the fetch response body. This is dependent on the implementation of the local transformer.

twifkak:
Enable recursive prefetch of script and style tags, by adding the
relevant Link header to the inner response headers of the SXG. This uses
the new Metadata proto returned by transformer.Process().

Fixes #37.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -276,21 +276,25 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
 	}
 }
 
-func formatLinkHeader(preloads []*rpb.Metadata_Preload) string {
+func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 	var values []string
 	for _, preload := range preloads {
-		var value strings.Builder
-		value.WriteByte('<')
 		u, err := url.Parse(preload.Url)
 		if err != nil {
-			log.Printf("Invalid preload URL: %q\n", preload.Url)
+			return "", errors.Wrapf(err, "Invalid preload URL: %q\n", preload.Url)
 		}
+		if preload.As == "" {
+			return "", errors.Errorf("Missing `as` attribute for preload URL: %q\n", preload.Url)
+		}
+
+		var value strings.Builder
+		value.WriteByte('<')
 		value.WriteString(u.String())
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -288,13 +306,16 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
...
 	r := getTransformerRequest(this.rtvCache, string(fetchBody), signURL.String())
-	transformed, _, err := transformer.Process(r)
+	transformed, metadata, err := transformer.Process(r)
 	if err != nil {
 		log.Println("Not packaging due to transformer error:", err)
 		proxy(resp, fetchResp, fetchBody)
 		return
 	}
 	fetchResp.Header.Set("Content-Length", strconv.Itoa(len(transformed)))
+	if linkHeader := formatLinkHeader(metadata.Preloads); linkHeader != "" {
```
honeybadgerdontcare: should there be some sanitizing/verifying of what to expect within metadata.Preloads here or within formatLinkHeader? it seems we're expecting Preloads to either be defined or empty and with a .Url (if not will fail parsing) and also a .As but we don't check for that to be the case.

what would happen if there is a preload.Url but not a preload.As?

packager/signer/signer.go
```diff
@@ -275,6 +276,23 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
...
 }
 
+func formatLinkHeader(preloads []*rpb.Metadata_Preload) string {
+	var values []string
+	for _, preload := range preloads {
+		var value strings.Builder
+		value.WriteByte('<')
+		u, err := url.Parse(preload.Url)
+		if err != nil {
+			log.Printf("Invalid preload URL: %q\n", preload.Url)
```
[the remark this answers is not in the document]
twifkak: Done and done.

packager/signer/signer.go
twifkak: OK, I added a basic check for As. I don't think there are any other checks to do? This comes from transformer.Process() anyway, so it shouldn't be structurally wrong.

PR #160, opened 2018-10-17
Limit the number of preloads to 20.

twifkak:
This caps the number of preloads in the Link header, so that prefetches
of the SXG have a bounded impact on network usage.


twifkak:
Actually, nevermind. I'm going to do this inside the transformer.

PR #167, opened 2018-10-22
Implement AMP-Cache-Transform version negotiation.

### LINKED ISSUE #29: Versioning
Implement versioning according to the [AMP-Cache-Transform spec](https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-transform.md).

twifkak:
Parse v parameters per the spec, and use the new
transform.SelectVersion() function to perform proactive version
negotiation given the request.

Fixes #29.

### DIFF
--- a/packager/amp_cache_transform/amp_cache_transform.go
+++ b/packager/amp_cache_transform/amp_cache_transform.go
@@ -39,6 +39,8 @@ func discardOWS(reader *strings.Reader) error {
 	return nil
 }
 
+const parameterisedListSeparator = ','
+
 // https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.3
 func parseParameterisedList(reader *strings.Reader) ([]parameterisedIdentifier, error) {
 	items := []parameterisedIdentifier{}
@@ -57,10 +59,10 @@ func parseParameterisedList(reader *strings.Reader) ([]parameterisedIdentifier, 
 		if err := discardOWS(reader); err != nil {
 			return nil, errors.Wrap(err, "discarding OWS")
 		}
-		if comma, err := reader.ReadByte(); err != nil {
-			return nil, errors.Wrap(err, "reading ','")
-		} else if comma != ',' {
-			return nil, errors.New("expected ','")
+		if char, err := reader.ReadByte(); err != nil {
+			return nil, errors.Wrapf(err, "reading '%c'", parameterisedListSeparator)
+		} else if char != parameterisedListSeparator {
+			return nil, errors.Errorf("expected '%c'", parameterisedListSeparator)
 		}
 		if err := discardOWS(reader); err != nil {
 			return nil, errors.Wrap(err, "discarding OWS")
@@ -71,6 +73,11 @@ func parseParameterisedList(reader *strings.Reader) ([]parameterisedIdentifier, 
 	}
 	return nil, errors.New("expected non-empty parameterised list")
 }
+
+const (
+	parameterSeparator = ';'
+	parameterValueSeparator = '='
+)
 
 // https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.4
 func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentifier, error) {
@@ -91,11 +98,11 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
 		if reader.Len() == 0 {
 			break
 		}
-		if semicolon, err := reader.ReadByte(); err != nil {
-			return nil, errors.Wrap(err, "reading ';'")
-		} else if semicolon != ';' {
+		if char, err := reader.ReadByte(); err != nil {
+			return nil, errors.Wrapf(err, "reading '%c'", parameterSeparator)
+		} else if char != parameterSeparator {
 			if err := reader.UnreadByte(); err != nil {
-				return nil, errors.Wrap(err, "unreading ';'")
+				return nil, errors.Wrapf(err, "unreading '%c'", parameterSeparator)
 			}
 			break
 		}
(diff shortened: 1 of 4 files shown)


packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -71,13 +80,102 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
...
+	for {
+		if err := discardOWS(reader); err != nil {
+			return nil, errors.Wrap(err, "discarding OWS")
+		}
+
+		// If we've reach the end of the parameter list, exit success.
+		if reader.Len() == 0 {
+			break
+		}
+		semicolon, err := reader.ReadByte()
```
alin04: Might be more readable to use the if initializer, e.g. 
if semicolor, err := reader.ReadByte(); err == nil {
    if semicolon != ';' {
    }
} else {
    return nil, errors.Wrap(....)
}

This keeps variable scope as minimal as possible - I don't see semicolon being referenced later.

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -71,13 +80,102 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
...
+		}
+		if name != "v" {
+			return nil, errors.Errorf("invalid AMP-Cache-Transform param %q", name)
+		}
+
+		// ... and a parameter value.
+		// NOTE: The current version of the AMP-Cache-Transform spec
+		// does not use any null-valued parameters. So, for now, a
+		// missing '=' is a parse failure.
+		equals, err := reader.ReadByte()
```
alin04: Same here for if initialization.

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -71,13 +80,102 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
...
+		}
+		if equals != '=' {
+			return nil, errors.New("expected '='")
+		}
+		// NOTE: In the future, this may need to be parseItem() to
+		// support other Item types:
+		// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.5
+		value, err := parseString(reader)
+		if err != nil {
+			return nil, errors.Wrapf(err, "parsing value for param %s", name)
```
alin04: Isn't name always 'v'? No need for the string format.

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -71,13 +80,102 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
...
+		}
+
+		params[name] = value
```
alin04: I think I've missed something. Isn't name always 'v'?

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -71,13 +80,102 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
...
+		if err != nil {
+			return "", errors.Wrap(err, "reading char")
+		}
+		if char <= 0x1f || char == 0x7f {
+			return "", errors.Errorf("invalid char %d", char)
+		}
+		if char == '"' {
+			break
+		}
+		if char == '\\' {
```
honeybadgerdontcare: shouldn't this be at line 161? according to 4.2.7 process point 4.2 happens before 4.4.

packager/amp_cache_transform/amp_cache_transform_test.go
```diff
@@ -3,30 +3,52 @@ package amp_cache_transform
...
+	"github.com/ampproject/amppackager/transformer"
+	rpb "github.com/ampproject/amppackager/transformer/request"
 	"github.com/stretchr/testify/assert"
 )
 
+func header(header string, version int64) string {
+	return header
+}
+
 func TestShouldSendSXG(t *testing.T) {
```
honeybadgerdontcare: Maybe add some that should break parsing so that it confirms the parsing is WAI.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Done.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Done.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: OK, I changed this. name can be anything now. Reason is that an invalid param should only fail one identifier, not the whole thing. (Added test.)

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Ditto.

packager/amp_cache_transform/amp_cache_transform.go
twifkak: Good catch. Yeah, I took a shortcut here moving the ifs around so that I could reuse the `value.WriteByte()` call below. It was (I think) equivalent behavior, but clearly not worth the confusion. Now it should read closer to the spec.

(Still a little different in the way it handles premature end-of-string: error-checking on the `ReadByte` instead of a condition on the `for` loop.)

packager/amp_cache_transform/amp_cache_transform_test.go
twifkak: Good idea. Added a bunch of different test cases, but I'm probably missing a bunch more. LMK if you have any in mind.

honeybadgerdontcare: Some additional ones for parseString:

```
`google;v="\`
`google;v="\a`
`google;v="\1`
```

I'd also test the condition `char <= 0x1f || char == 0x7f` but not exactly sure how to do that in a Go test.

twifkak: Done.

packager/amp_cache_transform/amp_cache_transform.go
```diff
@@ -71,13 +80,102 @@ func parseParameterisedIdentifier(reader *strings.Reader) (*parameterisedIdentif
...
+	return &parameterisedIdentifier{primaryID, params}, nil
+}
+
+// https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2.7
+func parseString(reader *strings.Reader) (string, error) {
+	quote, err := reader.ReadByte()
+	if err != nil {
+		return "", errors.Wrap(err, "reading '\"'")
+	}
+	if quote != '"' {
```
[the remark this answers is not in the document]
twifkak: Done.

PR #170, opened 2018-10-29
Make ServeHTTP responses Vary on Accept

ithinkihaveacat:
I think this is the right place to do it?

Ideally should also check that non-SXG responses also `Vary: Accept` but not sure where to do this.


packager/signer/signer.go
```diff
@@ -179,6 +179,8 @@ func (this *Signer) genCertURL(cert *x509.Certificate, signURL *url.URL) (*url.U
 }
 
 func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
+	resp.Header().Add("Vary", "Accept")
```
twifkak: While you're here, make it "Accept, AMP-Cache-Transform".

twifkak: I'm slightly partial to a single header `Vary: Accept, AMP-Cache-Transform` as it's slightly simpler (intermediaries are supposed to handle duplicate header names, but I wouldn't be surprised if some don't).

packager/signer/signer_test.go
```diff
@@ -155,6 +155,7 @@ func (this *SignerSuite) TestSimple() {
 	this.Assert().Equal(fmt.Sprintf(`google;v="%d"`, transformer.SupportedVersions[0].Max), resp.Header.Get("AMP-Cache-Transform"))
 	this.Assert().Equal("nosniff", resp.Header.Get("X-Content-Type-Options"))
 	this.Assert().Equal(fakePath, this.lastRequestURL)
+	this.Assert().Equal("Accept, AMP-Cache-Transform", resp.HeaderGet("Vary")
```
twifkak: Missing end paren. :)

ithinkihaveacat: Sigh. Fixed now!

twifkak:
Thanks, @ithinkihaveacat! Travis is happy, so I'm merging.

PR #171, opened 2018-10-31
Only forward requests if Accept includes a/s-e

ithinkihaveacat:
I don't think there's any point forwarding requests if the client hasn't indicated any desire to for an `application/signed-exchanged` response?

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,9 @@ For now, productionizing is a bit manual. The minimum steps are:
   3. Configure your TLS-serving frontend server to conditionally proxy to
      `amppkg`:
      1. If the URL starts with `/amppkg/`, forward the request unmodified.
-     2. If the URL points to an AMP page and the `AMP-Cache-Transform` request
-        header is present, rewrite the URL by prepending `/priv/doc` and forward
-        the request.
+     2. If the URL points to an AMP page, the `AMP-Cache-Transform` request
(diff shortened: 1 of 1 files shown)


twifkak:
Thanks, Michael! I didn't include this originally for a few reasons:

 1. It's hard to parse Accept headers correctly, and I'd rather not encourage the proliferation of inaccurate parsers.
 2. Nginx doesn't have an "&&" operator. The [workaround](http://rosslawley.co.uk/archive/old/2010/01/04/nginx-how-to-multiple-if-statements/) is a bit involved. Given people might want to make a config without that complexity, I wanted to make a recommendation that fits that case. (i.e. if you're going to condition on only one of the headers, it's better to do A-C-T, as eventually browsers will start sending Accept)
 3. There's (currently) no reason to send A-C-T without Accept: a/s-e. The A-C-T spec is all about requesting signed exchanges. So I think it signals intent well enough.
 4. amppkg will eventually check for Accept anyway, and proxy unsigned.

WDYT? You could maybe reword with an "if possible, also..." kinda thing. Or maybe conversely, "if you can only check for one, then..."

ithinkihaveacat:
Ok, thanks for the explanation! I agree with you that the front-end should just look for A-C-T, and adding in an "if possible…" to the README will just cause confusion. I'll close this.

PR #173, opened 2018-11-01
Add support for generating a fake OCSP response

alin04:
Add support for generating a fake OCSP response to seed the cache in development mode.


alin04:
I debated on replacing the fake ca.cert entirely, and regenerating the server.cert. Let me know your thoughts.

testdata/b1/README.md
```diff
@@ -14,6 +14,28 @@ $ openssl x509 -req -in server.csr -CA ca.cert -CAkey ca.privkey -CAcreateserial
 $ cat server.cert ca.cert >fullchain.cert
 ```
 
+OCSP:
+
+Make sure your openssl.cfg has an ocsp extension section, like:
+
+[ ocsp ]
```
twifkak: Put this and other code snippets in ``` quotes ``` like above. (Click View button to preview the markdown rendering.)

twifkak:
Yeah, if this new ca cert can be used in the same way as the old one, then go ahead and replace and regenerate. Would be nice to keep it simple if possible. Note you'll have to fix some unit tests because they hard-code the cert hash.

twifkak:
(OTOH LMK if there's some other reason not to replace & regenerate that you thought of.)

testdata/b1/README.md
alin04: Done

alin04:
I kept the ocsp cert separate due to problems when I tried to set the OCSPSigner extension on the root cert.

PR #181, opened 2018-11-02
Change example.com to amppackageexample.com.

### LINKED ISSUE #180: Update packager instructions with working test example
The packager instructions for testing the config don't work because example.com isn't AMP HTML.

The packager will skip packaging and proxy the content through with the error:
Not packaging due to transformer error: html tag is missing an AMP attribute

2018/11/02 13:45:35 Serving /priv/doc/https://example.com/ to [::1]:43942
2018/11/02 13:45:35 Fetching URL: "https://example.com/"
2018/11/02 13:45:35 No update necessary.
2018/11/02 13:45:35 Not packaging due to transformer error: html tag is missing an AMP attribute

Update with working example that will package and return a signed exchange.

twifkak:
This way the suggested minimal config will work in development mode, as
amppackageexample serves AMP. Also change the default OCSP cache file to
something that doesn't require a mkdir.

Fixes #180.


README.md
```diff
@@ -43,15 +43,14 @@ can obtain certificates for.
      PackagerBase = 'https://localhost:8080/'
      CertFile = 'path/to/fullchain.pem'
      KeyFile = 'path/to/privkey.pem'
-     OCSPCache = '/tmp/amppkg/ocsp'
+     OCSPCache = '/tmp/amppkg-ocsp'
```
alin04: Also update testdata/b1/README that has instructions on how to seed the cache. Its location needs to be changed too.

twifkak: Done. Also, it turns out I don't need to update the cert. `--ignore-certificate-errors-spki-list` also ignores the domain name being wrong. But I did anyway so it was consistent. (And found a few issues in doing so.)

Issue #162, opened 2018-10-18
Deployment


twifkak:
Publish some release resources (docs, scripts, containers, packages, whatever) useful enough for some production environments.

Gregable:
I think this is too vague as an issue, it's more of a broad theme. It's not clear what the deliverables would be to close this. Let's file issues for more specific stuff, like the Docker issue #182 if that's ok.

PR #182, opened 2018-11-02
Basic Docker support

alin04:
Very basic docker support, #162


docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN go get -v github.com/ampproject/amppackager/cmd/amppkg
+
+# Seed the ocsp cache
+WORKDIR /go/src/github.com/ampproject/amppackager/testdata/b1/
+RUN mkdir -p /tmp/amppkg && touch ./index.txt && touch ./index.txt.attr && openssl ocsp -index ./index.txt -rsigner ca.ocsp.cert -rkey ca.privkey -CA ca.cert -ndays 7 -reqin ocspreq.der -respout /tmp/amppkg/ocsp
+
+# Copy the binary and config file
+WORKDIR /go/src/app
+RUN cp /go/bin/amppkg .
+COPY . .
```
twifkak: It wasn't obvious to me that this copied the entire docker directory into the container. Perhaps mention that.

docker/amppkg.example.toml
```diff
@@ -0,0 +1,129 @@
+# Example toml using the fake testdata certs shipped with AMP Packager.
+#
+# The signed exchanges generated by this are NOT valid and will NOT work with
+# any browser or AMP cache!
+
+
+# This is a TOML 0.4.0 file, as specified by https://github.com/toml-lang/toml.
```
twifkak: Maybe just say "see ../amppkg.example.toml for details" instead of duplicating the comments here. Otherwise, they'll quickly get out of date.

docker/amppkg.example.toml
```diff
@@ -0,0 +1,129 @@
...
+  # the AMP Cache. By default, the URL that the frontend requests to sign is
+  # also the URL where the packager fetches it. For extra flexibility, see
+  # URLSet.Fetch below.
+  [URLSet.Sign]
+    # The scheme of the URL must be https. There is no way to configure this.
+    # The `user:pass@` portion is disallowed. There is no way to configure this.
+
+    # The domain to limit signed URLs to. An exact string match. The
+    # certificate must cover this domain.
+    Domain = "example.com"
```
twifkak: amppackageexample.com?

docker/amppkg.example.toml
```diff
@@ -0,0 +1,129 @@
...
+    # Examples of paths you might want to exclude:
+    #   * Personalized content, such as user settings pages. Signed exchanges
+    #     are cached globally and served to all users. (Personalization that
+    #     happens at runtime via JS is fine.) This provides defense-in-depth,
+    #     in addition to the Cache-Control: public check.
+    #   * User-generated content, such as forums. For instance, if there's a
+    #     chance of an XSS vulnerability in your templating library, the impact
+    #     of such event is higher here: even after you've fixed the bug, caches
+    #     may serve your signed packages for up to 7 days.
+    PathExcludeRE = []
```
twifkak: You can remove this or comment it out; it's the default. (I included the fix to the other example.toml in my PR to you.)

docker/amppkg.example.toml
```diff
@@ -0,0 +1,129 @@
...
+CertFile = '/go/src/github.com/ampproject/amppackager/testdata/b1/fullchain.cert'
+
+# The path to the PEM file containing the private key that corresponds to the
+# leaf certificate in CertFile.
+KeyFile = '/go/src/github.com/ampproject/amppackager/testdata/b1/server.privkey'
+
+# The path to a file where the OCSP response will be cached. The parent
+# directory should exist, but the file need not. If this is a network-mounted
+# file, it should support shared/exclusive locking.
+OCSPCache = '/tmp/amppkg/ocsp'
```
twifkak: Maybe make this `/tmp/amppkg-ocsp` to save the need for a `mkdir` in your Dockerfile.

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Use an official Go runtime as a parent image
+FROM golang:1.11
+
+# Install AMP Packager
+RUN go get -v github.com/ampproject/amppackager/cmd/amppkg
+
+# Seed the ocsp cache
+WORKDIR /go/src/github.com/ampproject/amppackager/testdata/b1/
+RUN mkdir -p /tmp/amppkg && touch ./index.txt && touch ./index.txt.attr && openssl ocsp -index ./index.txt -rsigner ca.ocsp.cert -rkey ca.privkey -CA ca.cert -ndays 7 -reqin ocspreq.der -respout /tmp/amppkg/ocsp
```
twifkak: My PR to you adds the empty `index.txt{,.attr}` files.

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Make port 8080 available to the world outside this container. This
+# port must match the AMP Packager port configured in the toml file.
+EXPOSE 8080
+
+# Start the AMP Packager
+ENTRYPOINT ["amppkg"]
+
+# Set default flags to run in development mode.
+CMD ["-development", "-config=amppkg.example.toml"]
```
ithinkihaveacat: `ARG` to specify the config might be useful here https://docs.docker.com/engine/reference/builder/#arg.

docker/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Run locally in development mode
+
+#### As-is
+
+To run the AMP Packager with the fake test certificates that it ships
+with (forwarding local port 8080 to the container's port 8080):
+
+```sh
+$ docker build -t amppackager
+$ docker run -p 8080:8080 amppackger
```
ithinkihaveacat: s/amppackger/amppackager/

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /go/src/app
+RUN cp /go/bin/amppkg .
+COPY . .
+
+# Make port 8080 available to the world outside this container. This
+# port must match the AMP Packager port configured in the toml file.
+EXPOSE 8080
+
+# Start the AMP Packager
+ENTRYPOINT ["amppkg"]
```
ithinkihaveacat: Is it possible to get a bash shell on the container with this `ENTRYPOINT`?

I'm not too familiar with Docker but if there's not `ENTRYPOINT`, and `CMD` is `["amppkg", "-development", "-config=amppkg.example.toml"]`, then `docker run -p 8080:8080 amppackager` will do the same as before, but it's also possible to do `docker run -it -p 8080:8080 amppackager bash` and get dropped into a bash shell on the image, for debugging purposes, etc.

docker/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Run locally in development mode
+
+#### As-is
+
+To run the AMP Packager with the fake test certificates that it ships
+with (forwarding local port 8080 to the container's port 8080):
+
+```sh
+$ docker build -t amppackager
```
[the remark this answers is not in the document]
alin04: Fixed. Will see in next push.

docker/README.md
alin04: Done

docker/amppkg.example.toml
alin04: Done (wait for next push).

docker/amppkg.example.toml
alin04: done

docker/amppkg.example.toml
alin04: removed.

docker/amppkg.example.toml
alin04: done

docker/Dockerfile
alin04: ack

alin04:
@ithinkihaveacat , no you don't need the ocsp cache since you're using real certs (that should have the OCSP URI encoded within). The fake certs that the AMP Packager ships with don't have that, so the binary won't start up (it blows up on CertCache.extractOCSPServer). 

Hence the hoop jumping to generate an OCSP response and seed the cache so the server will start.

docker/Dockerfile
alin04: Changed to only copy the config file (that points to the fake certs).

docker/Dockerfile
alin04: You are correct. I was leaning towards trying to keep the container more focused and obvious that it is an executable.

Entrypoint can be bypassed via 
$ docker run --entrypoint bash -it -p9999:8080 amppackager

WDYT?

alin04:
PTAL. I've addressed all comments with one open one still btw CMD vs ENTRYPOINT.

docker/Dockerfile
alin04: I'm a Docker newb, and I actually haven't tried, but I think ARG is only in scope during the build phase? Does that help when publishers/developers want to override it? Maybe ENV could work?

alin04:
I'm going to merge this in now. Can address the remaining docker questions as follow on.

PR #183, opened 2018-11-04
Support the packaging of ampbyexample.com [DO NOT MERGE]

ithinkihaveacat:
This PR add two features to support the packaging of `ampbyexample.com` as discussed in a separate PR in that repo (https://github.com/ampproject/amp-by-example/pull/1592).

The changes are:

1. Support for reading AMPs from the local filesystem (instead of over the network).
2. Support for filtering incoming requests via the `X-Appengine-Inbound-Appid` HTTP header.

Note that this PR is being used to provide a code review mechanism for the required changes; I don't think the changes are currently suitable to be merged into master.

/cc @sebastianbenz


cmd/amppkg/main.go
```diff
@@ -147,11 +164,19 @@ func main() {
...
+		log.Printf("X-Appengine-Inbound-Appid header must equal \"%s\"", config.ProjectId)
+	} else {
+		handler = logIntercept{mux}
+	}
 	server := http.Server{
 		Addr: addr,
 		// Don't use DefaultServeMux, per
 		// https://blog.cloudflare.com/exposing-go-on-the-internet/.
-		Handler:           logIntercept{mux},
+		// TODO Respond to /_ah/healthcheck with a 200
```
twifkak: May want to use certCache.IsHealthy for this.

cmd/amppkg/main.go
```diff
@@ -147,11 +164,19 @@ func main() {
 		addr = "localhost"
 	}
 	addr += fmt.Sprint(":", config.Port)
+	var handler http.Handler
```
twifkak: I'd dedupe with:

```
handler := logIntercept{mux}
if config.ProjectId != "" {
  handler = allowProject(config.ProjectId, handler)
  log.Printf(...)
}
```

packager/signer/signer.go
```diff
@@ -128,13 +128,25 @@ func noRedirects(req *http.Request, via []*http.Request) error {
...
 }
 
-func New(cert *x509.Certificate, key crypto.PrivateKey, urlSets []util.URLSet,
+func New(cert *x509.Certificate, key crypto.PrivateKey, docroot string, urlSets []util.URLSet,
 	rtvCache *rtv.RTVCache, shouldPackage func() bool, overrideBaseURL *url.URL,
 	requireHeaders bool) (*Signer, error) {
+	var t http.RoundTripper
+	if docroot != "" {
+		t := &http.Transport{}
+		r := http.NewFileTransport(http.Dir(docroot))
```
twifkak: Clever hack. :) TIL about NewFileTransport.

If you were merging this, I'd suggest maybe passing an `http.RoundTripper` into `signer.New` instead of a docroot. That separates some of the fetching logic from signer. (In honesty, probably more of the fetching logic should be extracted out of signer, but one step at a time.) We could even imagine, like, a SQL-backed RoundTripper or something. May eventually outgrow the RoundTripper interface.

packager/signer/signer.go
```diff
@@ -128,13 +128,25 @@ func noRedirects(req *http.Request, via []*http.Request) error {
...
+		t.RegisterProtocol("http", r)
```
twifkak: I think this is correct for the general case (where fetch and sign may differ). Just to clarify, though: [in your case](https://github.com/ampproject/amp-by-example/pull/1592/files#diff-95a5bd8f57d0cf05150691875cb9617b) (where fetch and sign are equal), it's not necessary, as the signed URL [must be https](https://wicg.github.io/webpackage/draft-yasskin-httpbis-origin-signed-exchanges-impl.html#application-signed-exchange).

cmd/amppkg/main.go
```diff
@@ -128,8 +145,8 @@ func main() {
 		}
 	}
 
-	packager, err := signer.New(certs[0], key, config.URLSet, rtvCache, certCache.IsHealthy,
-		overrideBaseURL, /*requireHeaders=*/!*flagDevelopment)
+	packager, err := signer.New(certs[0], key, config.Docroot, config.URLSet, rtvCache, certCache.IsHealthy,
+		overrideBaseURL /*requireHeaders=*/, !*flagDevelopment)
```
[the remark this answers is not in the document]
ithinkihaveacat: `gofmt` is responsible for shifting that comma, interestingly. (Also wants to reorder the imports a bit.)

cmd/amppkg/main.go
ithinkihaveacat: Thanks! I did try something like that originally but couldn't get it working. But, turns out that it needs to some type conversion to address the embedding:

```go
handler := http.Handler(logIntercept{mux})
...
```

packager/signer/signer.go
ithinkihaveacat: Yes, although perhaps the transport should be a function of the URLSet Domain? (So a different client per domain, in the same way there can be a different "fetch" per domain.)

twifkak: Yeah, good point.

twifkak:
Closing since I think discussion on this is done, but feel free to reopen. Thanks!

Issue #193, opened 2018-11-09
Obey HTTP requirements for gateways (reverse proxies)


twifkak:
Whether packaging as a signed exchange or proxying unsigned, the packager should strip [hop-by-hop headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#hbh) from the response. TODO: Find the section in RFC723[0-5] that specifies this, and find out if there's any other proxy requirements adjacent to that, that the packager isn't meeting.

twifkak:
OK, amppackager is a "gateway" per https://tools.ietf.org/html/rfc7230#section-2.3. Here's what a quick Ctrl-F for gateway turned up:

[3.2.4. Field Parsing](https://tools.ietf.org/html/rfc7230#section-3.2.4)
```
   A proxy or gateway that receives an obs-fold in a response message
   that is not within a message/http container MUST either discard the
   message and replace it with a 502 (Bad Gateway) response, preferably
   with a representation explaining that unacceptable line folding was
   received, or replace each received obs-fold with one or more SP
   octets prior to interpreting the field value or forwarding the
   message downstream.
```

[5.7.1. Via](https://tools.ietf.org/html/rfc7230#section-5.7.1)
```
   An HTTP-to-HTTP gateway
   MUST send an appropriate Via header field in each inbound request
   message and MAY send a Via header field in forwarded response
   messages.
```

[6.1. Connection](https://tools.ietf.org/html/rfc7230#section-6.1)
```
   The "Connection" header field allows the sender to indicate desired
   control options for the current connection.  In order to avoid
   confusing downstream recipients, a proxy or gateway MUST remove or
   replace any received connection options before forwarding the
   message.

   When a header field aside from Connection is used to supply control
   information for or about the current connection, the sender MUST list
   the corresponding field-name within the Connection header field.  A
   proxy or gateway MUST parse a received Connection header field before
   a message is forwarded and, for each connection-option in this field,
   remove any header field(s) from the message with the same name as the
   connection-option, and then remove the Connection header field itself
   (or replace it with the intermediary's own connection options for the
   forwarded message).
```

twifkak:
BTW, future versions of the SXG spec may mandate the Connection behavior per WICG/webpackage#336.

twifkak:
It seems like 3.2.4 is implemented by https://golang.org/pkg/net/textproto/#Reader.ReadMIMEHeader (via https://golang.org/src/net/http/response.go#L184).

[...]

PR #195, opened 2018-11-12
Remove Connection-specified headers.

twifkak:
Removes hop-by-hop headers such as Transfer-Encoding, so long as they
are specified by Connection. This should handle RFC7230-compliant
backend servers, but additional support for RFC2616 hop-by-hop header
removal may be warranted. More research is required.

Addresses #193.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


packager/signer/signer.go
```diff
@@ -89,6 +90,11 @@ var statusNotModifiedHeaders = map[string]bool{
 	"Vary":             true,
 }
 
+// A comma, as would appear in a Connection header. Comma-separation is defined
+// in https://tools.ietf.org/html/rfc7230#section-7, with OWS defined in
+// https://tools.ietf.org/html/rfc7230#appendix-B.
+var comma *regexp.Regexp = regexp.MustCompile(`[ \t]*,[ \t]*`)
```
alin04: Is this identical to the regex defined in accept.go? Consider consolidating in one place.

twifkak: Yeah, it's identical. Filed #196.

PR #210, opened 2018-12-07
Obey remaining HTTP requirements for gateways.

### LINKED ISSUE #193: Obey HTTP requirements for gateways (reverse proxies)

twifkak:
Remove hop-by-hop headers as specified in RFC 2616, and add Via header.

Fixes #193.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -115,6 +115,12 @@ var getTransformerRequest = func(r *rtv.RTVCache, s, u string) *rpb.Request {
 		AllowedFormats: []rpb.Request_HtmlFormat{rpb.Request_AMP}}
 }
 
+// Roughly matches the protocol grammar
+// (https://tools.ietf.org/html/rfc7230#section-6.7), which is defined in terms
+// of token (https://tools.ietf.org/html/rfc7230#section-3.2.6). This differs
+// in that it allows multiple slashes, as well as initial and terminal slashes.
+var protocol = regexp.MustCompile("^[!#$%&'*+\\-.^_`|~0-9a-zA-Z/]+$")
+
 // The following hop-by-hop headers should be removed even when not specified
 // in Connection, for backwards compatibility with downstream servers that were
 // written against RFC 2616, and expect gateways to behave according to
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -155,25 +180,23 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq http.Header) (*http.Re
 		return nil, nil, util.NewHTTPError(http.StatusInternalServerError, "Error building request: ", err)
 	}
 	req.Header.Set("User-Agent", userAgent)
+	// Set Via per https://tools.ietf.org/html/rfc7230#section-5.7.1.
+	via := strings.TrimPrefix(serveHTTPReq.Proto, "HTTP/") + " " + "amppkg"
```
alin04: Is this always uppercase HTTP?

twifkak: No, it depends on what gets sent by the client (i.e. the frontend server): https://golang.org/src/net/http/request.go#L909

But:
a) it's usually HTTP, and
b) protocol-name is optional per https://tools.ietf.org/html/rfc7230#section-5.7.1. It needs to be a token, per https://tools.ietf.org/html/rfc7230#section-6.7, and it looks like the golang parser is pretty lenient so it might not be if the client is doing something really weird (e.g. trying to attack), but I think the worst-cast scenario is that the Via header is wrong. Since the `line` that gets passed to `parseRequestLine` comes from `textproto.Reader`, I think it's already been sanitized of special header-field characters (so it can't "escape" from the headers into the body). Of course, even if it did, it's only the request body, which the client is free to construct as they wish.

That said, it took a lot of words there to explain why this isn't broken, which means it's brittle. I'll add a check that the Proto is a token. :)

twifkak: Done.

alin04: I was concerned with case sensitivity. Will it always be caps? I believe strings.Trim is case-sensitive?

twifkak: Yeah it's defined to be uppercase: https://tools.ietf.org/html/rfc7230#section-2.6

Broken clients can do broken things, but the TrimPrefix is an optimization anyways; it's allowed to specify the protocol-name.

PR #211, opened 2018-12-07
Disallow query strings by default.

### LINKED ISSUE #208: Change query-string regex in config to "" by default
This is a safer default, per WICG/webpackage#321. It's a backwards-incompatible change -- a luxury we can afford only during developer preview -- it'd be nice to double-check no production servers are depending on it.

twifkak:
This is a change to the default behavior of QueryRE in the config file;
users needing to sign URLs with query parameters will need to update
their configs.

Fixes #208.

### DIFF
--- a/packager/util/config.go
+++ b/packager/util/config.go
@@ -48,6 +48,7 @@ type URLPattern struct {
 	SamePath               *bool
 }
 
+var emptyRegexp = ""
 var dotStarRegexp = ".*"
 
 // Also sets defaults.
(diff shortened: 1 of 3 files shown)


packager/util/config.go
```diff
@@ -48,6 +48,7 @@ type URLPattern struct {
 	SamePath               *bool
 }
 
+var emptyRegexp = ""
 var dotStarRegexp = ".*"
```
honeybadgerdontcare: any reason to keep `dotStarRegexp`?

twifkak: Yeah, still used above for PathRE. (For now.)

PR #214, opened 2018-12-10
Add a 2000-char URL length limit by default.

twifkak:
This is to mitigate the content-sniffing risk outlined in
WICG/webpackage#348.

cc @molnarg

### DIFF
--- a/packager/util/config.go
+++ b/packager/util/config.go
@@ -49,7 +49,7 @@ type URLPattern struct {
 }
 
 var emptyRegexp = ""
-var defaultPathRegexp = ".{,1000}"
+var defaultPathRegexp = ".{,2000}"
 
 // Also sets defaults.
 func validateURLPattern(pattern *URLPattern) error {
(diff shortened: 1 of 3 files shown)


packager/util/config.go
```diff
@@ -49,12 +49,12 @@ type URLPattern struct {
 }
 
 var emptyRegexp = ""
-var dotStarRegexp = ".*"
+var defaultPathRegexp = ".{,1000}"
```
alin04: Is 1000 arbitrary? I didn't see any specific guidance by WICG?

alin04: Also, this unfairly dings encoded chars, e.g. %20 over +. Not sure if that's important.

twifkak: This is arbitrary. Just upped it to 2000 which is a bit of a de facto limit anyway per https://stackoverflow.com/a/417184/1440915.

True about pct-encoded chars. The `regexpFullMatch` is against `url.EscapedPath()`. I think that's the right thing to do? It makes the URL more "canonical" -- if we use `url.Path` then the client may or may not pct-encode their URL beforehand, which makes it harder to write a regex that covers both situations. So I'd say this unfairness is an acceptable side effect.

PR #216, opened 2018-12-13
Basic test benchmark for WebAssembly integration.

twifkak:
This is not nearly as efficient as a native build of the transform
binary, due to bugs such as http://crbug.com/853685 and
https://github.com/golang/go/issues/27462, but serves as a starting
point for further investigation.

There have been a few optimizations already:
 - Keeping one long-lived process open to amortize the bootstrapping
   cost across requests.
 - Batching the "requests" to the wasm code via Promise.all. (This can
   be seen by adding a log statement just before the call to exports.run
   in wasm_exec.js.)

Another potential problem to investigate the cost of is the lack of
cross-heap GC.

### DIFF
--- a/cmd/transform_wasm/main.go
+++ b/cmd/transform_wasm/main.go
@@ -1,0 +1,43 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


twifkak:
internal bug http://b/119071852

PR #245, opened 2019-02-13
Content-Security-Policy

Gregable:
Allow the AMP Packager to maintain some of the publisher's own
Content-Security-Policy directives, while ensuring that the final
Content-Security-Policy header is set to a value that will not break
AMP pages served on the AMP Cache.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
 	return req, resp, nil
 }
 
+// Some Content-Security-Policy (CSP) configurations have the ability to break
+// AMPHTML document functionality on the AMPHTML Cache if set on the document.
+// This method parses the publisher's provided CSP and mutates it to ensure
+// that the document is not broken on the AMP Cache.
+//
+// Specifically, the following CSP directives are passed through unmodified:
+//  - base-uri
+//  - block-all-mixed-content
+//  - font-src
+//  - form-action
+//  - manifest-src
+//  - referrer
+//  - upgrade-insecure-requests
+// And the following CSP directives are overridden to specific values:
+//  - object-src
+//  - report-uri
+//  - script-src
+//  - style-src
+//  - default-src
+// All other CSP directives are stripped from the publisher provided CSP.
+func MutateFetchedContentSecurityPolicy(fetched string) (string) {
+	directiveTokens := strings.Split(fetched, ";")
+	var newCsp string
+	for i := range directiveTokens {
+		directiveToken := directiveTokens[i]
+		trimmed := strings.TrimSpace(directiveToken)
+		directiveParts := strings.Fields(trimmed)
+		if len(directiveParts) == 0 {
+			continue
+		}
+		directiveName := strings.ToLower(directiveParts[0])
+		switch directiveName {
+			// Preserve certain directives. The rest are all removed or replaced.
+			case "base-uri", "block-all-mixed-content", "font-src", "form-action",
+			     "manifest-src", "referrer", "upgrade-insecure-requests":
+					 newCsp += trimmed + ";"
+			default:
+		}
+	}
+	// Add missing directives or replace the ones that were removed in some cases
+	newCsp += "default-src * blob: data:;" +
+	          "report-uri https://csp-collector.appspot.com/csp/amp;" +
+						"script-src blob: https://cdn.ampproject.org/rtv/ " +
+              "https://cdn.ampproject.org/v0.js " + 
+							"https://cdn.ampproject.org/v0/ " +
+              "https://cdn.ampproject.org/viewer/;" +
+            "style-src 'unsafe-inline' https://cdn.ampproject.org/rtv/ " +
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
...
+//  - manifest-src
+//  - referrer
+//  - upgrade-insecure-requests
+// And the following CSP directives are overridden to specific values:
+//  - object-src
+//  - report-uri
+//  - script-src
+//  - style-src
+//  - default-src
+// All other CSP directives are stripped from the publisher provided CSP.
```
twifkak: Add a link to https://w3c.github.io/webappsec-csp/ here.

packager/signer/signer.go
```diff
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
...
+func MutateFetchedContentSecurityPolicy(fetched string) (string) {
+	directiveTokens := strings.Split(fetched, ";")
+	var newCsp string
+	for i := range directiveTokens {
```
twifkak: Say `for _, directiveToken := range directiveTokens {` instead.

packager/signer/signer.go
```diff
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
...
+// And the following CSP directives are overridden to specific values:
+//  - object-src
+//  - report-uri
+//  - script-src
+//  - style-src
+//  - default-src
+// All other CSP directives are stripped from the publisher provided CSP.
+func MutateFetchedContentSecurityPolicy(fetched string) (string) {
+	directiveTokens := strings.Split(fetched, ";")
+	var newCsp string
```
twifkak: Use https://golang.org/pkg/strings/#Builder instead.

packager/signer/signer.go
```diff
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
...
+	for i := range directiveTokens {
+		directiveToken := directiveTokens[i]
+		trimmed := strings.TrimSpace(directiveToken)
+		directiveParts := strings.Fields(trimmed)
```
twifkak: Add a note that this differs slightly from the spec in that it includes `U+000b vertical tab` in its definition of white space.

(Sources: https://infra.spec.whatwg.org/#ascii-whitespace vs https://golang.org/pkg/strings/#Fields, https://golang.org/pkg/unicode/#IsSpace, and http://xahlee.info/golang/golang_string_backslash_escape.html)

packager/signer/signer.go
```diff
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
...
+							"https://cdn.ampproject.org/v0/ " +
+              "https://cdn.ampproject.org/viewer/;" +
+            "style-src 'unsafe-inline' https://cdn.ampproject.org/rtv/ " +
+              "https://cdn.materialdesignicons.com " +
+              "https://cloud.typography.com https://fast.fonts.net " + 
+              "https://fonts.googleapis.com https://maxcdn.bootstrapcdn.com " +
+              "https://p.typekit.net https://pro.fontawesome.com " +
+              "https://use.fontawesome.com https://use.typekit.net;" +
+            "object-src 'none'"
+	return newCsp
```
twifkak: Is the default CSP returned by this function (if the pub doesn't specify one) byte-exact to our current CSP? If not, we should probably disable this until the server-side change is complete.

packager/signer/signer.go
```diff
@@ -219,6 +219,62 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
...
+              "https://cdn.ampproject.org/v0.js " + 
+							"https://cdn.ampproject.org/v0/ " +
+              "https://cdn.ampproject.org/viewer/;" +
+            "style-src 'unsafe-inline' https://cdn.ampproject.org/rtv/ " +
+              "https://cdn.materialdesignicons.com " +
+              "https://cloud.typography.com https://fast.fonts.net " + 
+              "https://fonts.googleapis.com https://maxcdn.bootstrapcdn.com " +
+              "https://p.typekit.net https://pro.fontawesome.com " +
+              "https://use.fontawesome.com https://use.typekit.net;" +
+            "object-src 'none'"
```
twifkak: That is some wacky indenting right there, at least with GitHub's tabstop. Can you choose a more boring layout? e.g. Whatever `go fmt` produces.

packager/signer/signer.go
Gregable: No. However, rather than disabling/enabling, I'll just sequence the commits correctly.

packager/signer/signer.go
Gregable: one person's "wacky" is another person's "readable", but `go fmt` wins in the end.

Gregable:
No worries on the delay. Will wait to commit until the AMP Cache will not break on these.

PR #256, opened 2019-03-06
Rewrite host to that of signURL when fetchURL is different.

shigeki:
The `Host` header in requesting to fetchURL needs to be the same as that to signURL even when fetchURL is different from signURL.  This is the similar behaviors such as using a reverse proxy.

go fmt and go test are fine as below.
```
$ go fmt packager/signer/signer.go
$ go fmt packager/signer/signer_test.go
$ go test ./...
?       github.com/ampproject/amppackager/cmd/amppkg    [no test files]
?       github.com/ampproject/amppackager/cmd/amppkg_dl_sxg     [no test files]
?       github.com/ampproject/amppackager/cmd/amppkg_test_cache [no test files]
?       github.com/ampproject/amppackager/cmd/transform [no test files]
ok      github.com/ampproject/amppackager/packager/accept       (cached)
ok      github.com/ampproject/amppackager/packager/amp_cache_transform  (cached)
ok      github.com/ampproject/amppackager/packager/certcache    (cached)
ok      github.com/ampproject/amppackager/packager/rtv  (cached)
ok      github.com/ampproject/amppackager/packager/signer       (cached)
?       github.com/ampproject/amppackager/packager/testing      [no test files]
ok      github.com/ampproject/amppackager/packager/util (cached)
ok      github.com/ampproject/amppackager/packager/validitymap  (cached)
ok      github.com/ampproject/amppackager/transformer   (cached)
ok      github.com/ampproject/amppackager/transformer/internal/amphtml  (cached)
ok      github.com/ampproject/amppackager/transformer/internal/css      (cached)
ok      github.com/ampproject/amppackager/transformer/internal/htmlnode (cached)
?       github.com/ampproject/amppackager/transformer/internal/testing  [no test files]
ok      github.com/ampproject/amppackager/transformer/layout    (cached)
ok      github.com/ampproject/amppackager/transformer/printer   (cached)
?       github.com/ampproject/amppackager/transformer/request   [no test files]
ok      github.com/ampproject/amppackager/transformer/transformers      (cached)
```

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -192,7 +192,7 @@ func New(cert *x509.Certificate, key crypto.PrivateKey, urlSets []util.URLSet,
 	return &Signer{cert, key, &client, urlSets, rtvCache, shouldPackage, overrideBaseURL, requireHeaders}, nil
 }
 
-func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.Request, *http.Response, *util.HTTPError) {
+func (this *Signer) fetchURL(fetch *url.URL, sign *url.URL, serveHTTPReq *http.Request) (*http.Request, *http.Response, *util.HTTPError) {
 	ampURL := fetch.String()
 
 	log.Printf("Fetching URL: %q\n", ampURL)
@@ -216,6 +216,10 @@ func (this *Signer) fetchURL(fetch *url.URL, serveHTTPReq *http.Request) (*http.
 		if value := GetJoined(serveHTTPReq.Header, header); value != "" {
 			req.Header.Set(header, value)
 		}
+	}
+	// Rewrite host to that of signURL when fetchURL is different.
+	if req.Host != sign.Host {
+		req.Host = sign.Host
 	}
 	resp, err := this.client.Do(req)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


shigeki:
Close this for this has a duplicated feature with #259 .

twifkak:
Okay, sounds good. Sorry for the delay -- my initial thoughts were this doesn't seem right when `fetch.Path != sign.Path`, but I wanted to do a bit more research on what other gateways do in the presence of URL rewriting, before making a suggestion.

I'd also note that, if I'm reading it correctly, https://tools.ietf.org/html/rfc7230#section-5.4 doesn't mention anything about reverse proxies (aka gateways); only forward proxies. I think that's why https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded exist.

Earlier, I [decided against](https://github.com/ampproject/amppackager/issues/193#issuecomment-445084264) setting these headers because I was worried about the same over-signing concern -- I would prefer a default that doesn't have obvious avenues for user identification. However, in retrospect, I think that may have been too broad a decision. Including `X-Forwarded-Host` and/or `Forwarded: host=...` in the fetch request seems appropriate. I'll file a bug for that.

twifkak:
Okay, filed #260.

PR #264, opened 2019-03-13
Add gateway server code and updates protobuf package.

amaltas:
Creating a new pull request in a brand new client.


cmd/gateway_server/server.go
```diff
@@ -0,0 +1,190 @@
...
+		},
+	}
+
+	packager, err := signer.New(certs[0], privateKey, urlSets, rtvCache, shouldPackage, signUrl, false)
+
+	if err != nil {
+		return errorToSXGResponse(err), nil
+	}
+
+	if packager == nil {
```
twifkak: I think it's safe to get rid of this check. Though it's not documented, the convention for `New()` is to return a non-nil error if you plan on returning a nil value.

cmd/gateway_server/server.go
```diff
@@ -0,0 +1,190 @@
...
+		return errorToSXGResponse(err), nil
```
twifkak: But if you do keep it around, replace `err` here with a hard-coded string, for a more useful SXGResponse.

PR #266, opened 2019-03-20
Don't sign versioned Variants headers.

twifkak:
These are the ones that Chromium actually uses at the moment, per
https://crrev.com/86e0430c.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -398,7 +398,9 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
 
 		fetchResp.Header.Del("Link") // Ensure there are no privacy-violating Link:rel=preload headers.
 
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -398,7 +398,9 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
 
 		fetchResp.Header.Del("Link") // Ensure there are no privacy-violating Link:rel=preload headers.
 
-		if fetchResp.Header.Get("Variants") != "" || fetchResp.Header.Get("Variant-Key") != "" {
+		if fetchResp.Header.Get("Variants") != "" || fetchResp.Header.Get("Variant-Key") != "" ||
+			// Include versioned headers per https://github.com/WICG/webpackage/pull/406.
+			fetchResp.Header.Get("Variants-04") != "" || fetchResp.Header.Get("Variant-Key-04") != "" {
```
alin04: Add a test?

Issue #188, opened 2018-11-07
Add -staging flag to host cert-url elsewhere


twifkak:
When testing amppackager on a staging server (using the real certificate for X.com but hosting it on staging.X.com), one of the problems is that the cert-url is on X.com. Support an override for this, probably via flag since it's shouldn't be a permanent config in production. e.g. `-staging staging.X.com` will override the cert-url base, much like `-development` overrides it with localhost.

cc @ithinkihaveacat

twifkak:
Look at `overrideBaseURL` in `cmd/amppkg/main.go`.

PR #278, opened 2019-04-10
Support new -invalidcert flag

mattwomple:
Run amppackager in production mode with known invalid SSL certificates.
Useful to emulate an otherwise fully production environment without
having to pay the sole provider of SSL certificates with
CanSignHttpExchanges extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ packager URL directly, first add a Chrome extension to send an
 
 It is possible to test an otherwise fully production configuration without
 obtaining a certificate with the `CanSignHttpExchanges` extension. `amppkg`
-still needs to perform OCSP verification, so the Root CA must be valid (i.e. no
+still needs to perform OCSP verification, so the Issuer CA must be valid (i.e. no
 self-signed certificates). e.g. You can use a certificate from [Let's Encrypt](https://letsencrypt.org/).
 
 Running `amppkg` with the `-invalidcert` flag will skip the check for
--- a/README.md
+++ b/README.md
@@ -150,6 +150,21 @@ packager URL directly, first add a Chrome extension to send an
 `AMP-Cache-Transform: any` request header. Otherwise, follow the above
 "Demonstrate privacy-preserving prefetch" instructions.
 
+#### Testing productionization without a valid certificate
+
+It is possible to test an otherwise fully production configuration without
+obtaining a certificate with the `CanSignHttpExchanges` extension. `amppkg`
+still needs to perform OCSP verification, so the Root CA must be valid (i.e. no
+self-signed certificates). e.g. You can use a certificate from [Let's Encrypt](https://letsencrypt.org/).
+
+Running `amppkg` with the `-invalidcert` flag will skip the check for
+`CanSignHttpExchanges`. This flag is not necessary when using the
+`-development` flag.
+
+Chrome can be configured to allow these invalid certificates with the
+*Allow Signed HTTP Exchange certificates without extension* experiment:
+chrome://flags/#allow-sxg-certs-without-extension
+
 #### Redundancy
 
 If you need to load balance across multiple instances of `amppkg`, you'll want
--- a/cmd/amppkg/main.go
+++ b/cmd/amppkg/main.go
@@ -127,7 +127,7 @@ func main() {
 	}
 
 	packager, err := signer.New(certs[0], key, config.URLSet, rtvCache, certCache.IsHealthy,
-		overrideBaseURL, /*requireHeaders=*/!(*flagDevelopment || *flagInvalidCert))
+		overrideBaseURL, /*requireHeaders=*/!*flagDevelopment)
 	if err != nil {
 		die(errors.Wrap(err, "building packager"))
 	}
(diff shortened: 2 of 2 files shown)


mattwomple:
@Gregable I'm not normally in the camp of *we need more flags!*, but this feature would have been very useful to me while I was initially testing amppackager. Not a problem if you would rather implement this differently; I just hope it can inspire thoughts about making first-time use of amppackager easier.

README.md
```diff
@@ -150,6 +150,21 @@ packager URL directly, first add a Chrome extension to send an
 `AMP-Cache-Transform: any` request header. Otherwise, follow the above
 "Demonstrate privacy-preserving prefetch" instructions.
 
+#### Testing productionization without a valid certificate
+
+It is possible to test an otherwise fully production configuration without
+obtaining a certificate with the `CanSignHttpExchanges` extension. `amppkg`
+still needs to perform OCSP verification, so the Root CA must be valid (i.e. no
```
twifkak: s/Root/Issuer/ would be more accurate, I think.

twifkak:
Thanks! I'm generally in the same camp as you, but I'm okay with flags for things that signal this isn't a production instance. (It's easier to accidentally leave something in a config file than to accidentally paste it into your init script.)

FYI there's a similar FR at https://github.com/ampproject/amppackager/issues/188. That one addresses a different stage: I've got a production SXG cert from a CA, but I want to test it before intercepting production requests. I think your PR can coexist with that one—they probably should be separate flags—but I wanted to point it out in case you had ideas.

cmd/amppkg/main.go
```diff
@@ -90,7 +91,7 @@ func main() {
 	if certs == nil || len(certs) == 0 {
 		die(fmt.Sprintf("no cert found in %s", config.CertFile))
 	}
-	if !*flagDevelopment && !util.CanSignHttpExchanges(certs[0]) {
+	if (!*flagDevelopment && !*flagInvalidCert) && !util.CanSignHttpExchanges(certs[0]) {
```
twifkak: I slightly prefer the De Morgan equivalent `!(*flagDevelopment || *flagInvalidCert)` but am amenable to either.

mattwomple: No problem. Does this apply to line 130 as well? e.g.
`/*requireHeaders=*/!(*flagDevelopment || *flagInvalidCert)`

twifkak: Yeah.

README.md
mattwomple: Resolved in commit https://github.com/ampproject/amppackager/pull/278/commits/87ce94ae49b53beb76a96eff6b02e94ab515534d

cmd/amppkg/main.go
mattwomple: Resolved in commit f78e86e

cmd/amppkg/main.go
```diff
@@ -91,7 +91,7 @@ func main() {
 	if certs == nil || len(certs) == 0 {
 		die(fmt.Sprintf("no cert found in %s", config.CertFile))
 	}
-	if (!*flagDevelopment && !*flagInvalidCert) && !util.CanSignHttpExchanges(certs[0]) {
+	if !(*flagDevelopment || *flagInvalidCert) && !util.CanSignHttpExchanges(certs[0]) {
```
twifkak: Oops, actually it reads even better if you De Morgan again:
`!(*flagDevelopment || *flagInvalidCert || util.CanSignHttpExchanges(certs[0]))`

But that's outside the scope of your change, so feel free to ignore.

cmd/amppkg/main.go
```diff
@@ -127,7 +127,7 @@ func main() {
 	}
 
 	packager, err := signer.New(certs[0], key, config.URLSet, rtvCache, certCache.IsHealthy,
-		overrideBaseURL, /*requireHeaders=*/!*flagDevelopment && !*flagInvalidCert)
+		overrideBaseURL, /*requireHeaders=*/!(*flagDevelopment || *flagInvalidCert))
```
twifkak: Actually, on further consideration, are you sure `flagInvalidCert` belongs here? `requireHeaders` seems unrelated to the invalid cert:
https://github.com/ampproject/amppackager/blob/e4bf0430ba152cfe82ccf063df92021dfc0f26a5/packager/signer/signer.go#L352-L362

For testing one can use `curl`, `amppkg_dl_sxg`, or something like [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj) or [Requestly](https://www.requestly.in/).

mattwomple: Thanks for looking at this closer. You're correct. The `-invalidcert` flag is meant to be "otherwise production", so yes, we should expect a valid `AMP-Cache-Transform` header.

mattwomple: I verified commit 97f8a38 works as intended when sending values `nothing` and `any` for header `AMP-Cache-Transform`.

cmd/amppkg/main.go
mattwomple: Resolved in commit 79453d9

twifkak:
Done. Note this is `master` while the default branch is `releases`, so `go get` won't pick this up until we do a release. Can `git clone -b` in the meantime.

PR #283, opened 2019-04-24
certcache: Improve compatibility with Windows

### LINKED ISSUE #282: OCSP cache file creation causes amppkg to panic on Windows
On Windows 10 (build 17763) and compiling with `go version` go1.12.2 windows/amd64, amppkg panics during OCSP cache file creation. The current `Read` method in `packager/certcache/storage.go` only succeeds if a valid OCSP cache file already exists. It appears there's both problems with calling `ioutil.ReadFile(this.path)` as well as attempting to obtain an exclusive lock with `lock.TryLock()`. I've written a workaround to create the file if it does not yet exist, but it's not yet what I would consider an acceptable patch: https://github.com/mattwomple/amppackager/commit/5bca5e07d2d2ace4b8ad076b54d430cf6defe943 . Based on comments in [this](https://stackoverflow.com/questions/35558787/create-an-empty-text-file/35558965) StackOverflow question, it should not be necessary to check for existence before creating the file. Yet, in the above patch, if either of the commented sections _Create empty OCSP cache file._ or _Obtain an exclusive lock._ are uncommented, the panic returns. Do you have any suggestions for a proper fix?

For completeness, below is the stack trace output to console, but it's not useful:

    2019/04/18 08:03:28 Updating OCSP; none cached yet.
    2019/04/18 08:03:28 Updating OCSP; none cached yet.
    2019/04/18 08:03:28 Missing OCSP response.
    github.com/ampproject/amppackager/packager/certcache.(*CertCache).readOCSP
            C:/Users/mdmower/go/src/github.com/ampproject/amppackager/packager/certcache/certcache.go:229
    github.com/ampproject/amppackager/packager/certcache.(*CertCache).Init
            C:/Users/mdmower/go/src/github.com/ampproject/amppackager/packager/certcache/certcache.go:108
    main.main
            C:/Users/mdmower/go/src/github.com/ampproject/amppackager/cmd/amppkg/main.go:111
    runtime.main
            C:/Users/mdmower/Programs/Go/src/runtime/proc.go:200
    runtime.goexit
            C:/Users/mdmower/Programs/Go/src/runtime/asm_amd64.s:1337
    initializing CertCache
    github.com/ampproject/amppackager/packager/certcache.(*CertCache).Init
            C:/Users/mdmower/go/src/github.com/ampproject/amppackager/packager/certcache/certcache.go:110
    main.main
            C:/Users/mdmower/go/src/github.com/ampproject/amppackager/cmd/amppkg/main.go:111
    runtime.main
            C:/Users/mdmower/Programs/Go/src/runtime/proc.go:200
    runtime.goexit
            C:/Users/mdmower/Programs/Go/src/runtime/asm_amd64.s:1337
    building cert cache
    main.main
            C:/Users/mdmower/go/src/github.com/ampproject/amppackager/cmd/amppkg/main.go:112
    runtime.main
            C:/Users/mdmower/Programs/Go/src/runtime/proc.go:200
    runtime.goexit
            C:/Users/mdmower/Programs/Go/src/runtime/asm_amd64.s:1337

Finally, please avoid judging me for testing on Windows... as the old saying goes, "I'm asking for a friend." But really, I am!

mattwomple:
Windows does not allow read/write operations on a file with an exclusive
lock, "The process cannot access the file because another process has
locked a portion of the file." Use an independent lock file for better
cross-platform compatibility.

Windows also does not succeed at creating a new cache file with
ioutil.ReadFile(), "The system cannot find the file specified." Check
for file existence before attempting read operations.

flock is unable to transition a shared lock to an exclusive lock on
Windows. Unlock before locking again, but only on Windows.

Fixes #282


packager/certcache/storage.go
```diff
@@ -71,23 +105,33 @@ func (this *LocalFile) Read(ctx context.Context, isExpired func([]byte) bool, up
 			return contents, nil
 		}
 		// Upgrade to a write-lock. It seems this may or may not be atomic, depending on the system.
+		// Windows does not handle a lock "upgrade", hence unlock before lock.
```
twifkak: This looks good to me, but let me think out loud so you can double-check my work:

I guess there's a risk of a race in the gap between the Unlock and TryLock, but as I noted on the line above, that risk probably already existed. And I think it's not a problem since it's okay to grab two OCSP responses and write them in succession. (Atomic read-modify-write isn't strictly required here; just an optimization against hammering the OCSP responder from all N instances at once. On further thought, it may be a good idea to add some randomness to the ocspCheckInterval in certcache.go, just to avoid hammering the lock file.)

No risk of deadlock, since only one lock file is in use. (I could be wrong; I often am when it comes to concurrency.)

packager/certcache/storage.go
```diff
@@ -41,28 +43,60 @@ type LocalFile struct {
...
+		return false, nil
+	}
+	return true, err
+}
+
 func (this *LocalFile) Read(ctx context.Context, isExpired func([]byte) bool, update func([]byte) []byte) ([]byte, error) {
-	lock := flock.NewFlock(this.path)
+	// Use independent .lock file; necessary on Windows to avoid "The process cannot
+	// access the file because another process has locked a portion of the file."
+	lockPath := this.path + ".lock"
```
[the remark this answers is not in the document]
mattwomple: Addressed in commit deb6aea9085f35e72d90c77e4b5e28ee38582a8e

packager/certcache/storage.go
mattwomple: This is certainly my least favorite part of this commit series. As you noted, though: _that risk probably already existed_. Limiting this change to Windows means we aren't making things worse, as the attempts to transition lock type were causing `locked` to be false and so this process _would_ exit.

_No risk of deadlock, since only one lock file is in use_ : this thought occurred to me as well, since we will soon test running multiple instances of amppkg with a shared OCSP cache file. It is part of the reason I opted for hard-coding the `__.lock` filename instead of allowing it to be configurable.

All said, as much as I don't _like_ many of the changes in this commit series, I do think they introduce better compatibility without sacrificing anything on non-Windows systems.

mattwomple:
> Please run `dep ensure` (we're using https://golang.github.io/dep/ for now) to update the vendor/ directory.

These changes have been backed-out since they were unrelated to the pull request topic.

packager/certcache/storage.go
twifkak: Yep, I'm happy that amppkg is gaining support for Windows (no judgment :). In general, I'd love to minimize complexity, but not at the cost of correctness. Thanks for confirming. And, ugh, sorry for the overly-complicated certcache code. To quote [someone](https://twitter.com/cramforce/status/949054860166553600), "I was young and needed to ship."

twifkak:
... and you'll need to rebase and resolve since I merged #287.

mattwomple:
> ... and you'll need to rebase and resolve since I merged #287.

What's the preferred strategy? I can rebase (and optionally squash the commits), then force push. Or the current version of this pull request uses merge strategy, and I can leave it alone.

twifkak:
Oh, sorry, I missed your merge commit. That's fine, since I'll just squash.

PR #287, opened 2019-04-26
certcache: Switch to gofrs/flock

mattwomple:
Update the include to reference github.com/gofrs/flock which is the new
home of github.com/theckman/go-flock. Also, update to flock.New() which
is now preferred over flock.NewFlock().


mattwomple:
This will conflict temporarily with #283 but I'll resolve that conflict when (if) this or that is merged.

PR #289, opened 2019-04-29
Support HEAD requests

### LINKED ISSUE #261: Support HEAD requests
signer.go currently returns a 405 for HEAD requests. It should instead perform the same logic as for GET, to determine if it would return an `application/signed-exchange` or whatever the underlying fetch content-type is.

(Placing this in the v3 milestone only because it should be quick to do; otherwise it's not a super-high priority.)

mattwomple:
Perform identical operations as used in GET requests for doc, cert, and
validity. Rely on the http server to blank the body before responding.

Resolves #261

### DIFF
--- a/cmd/amppkg/main.go
+++ b/cmd/amppkg/main.go
@@ -136,14 +136,23 @@ func main() {
 	mux := httprouter.New()
 	mux.RedirectTrailingSlash = false
 	mux.RedirectFixedPath = false
-	mux.GET(util.ValidityMapPath, validityMap.ServeHTTP)
-	mux.HEAD(util.ValidityMapPath, validityMap.ServeHTTP)
-	mux.GET("/priv/doc", packager.ServeHTTP)
-	mux.HEAD("/priv/doc", packager.ServeHTTP)
-	mux.GET("/priv/doc/*signURL", packager.ServeHTTP)
-	mux.HEAD("/priv/doc/*signURL", packager.ServeHTTP)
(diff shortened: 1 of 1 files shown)


mattwomple:
@twifkak This is the bare-minimum approach, but it does seem to work fine. It could potentially be optimized by preventing the actual packaging, signing, etc that generates a response body, but that comes at the cost of complexity. Let me know your thoughts.

twifkak:
Thanks! I'm fine with the performance cost of all the signing, etc., as I expect this not to be a significant percentage of requests (correct me if you think otherwise).

Just to confirm: This doesn't actually cause a response body to be emitted, right? I'm guessing from 
https://golang.org/src/net/http/server.go#L372 that this is the case, but I imagine you've tested.

twifkak:
Optionally, make a little helper function like `func headAndGet(mux, path, handler)` or so. Or maybe make a loop like:

```
handlers := []struct{path string, handler *http.Handler}{
  ...
}
for _, handler := range handlers {
  mux.GET(...)
  mux.HEAD(...)
}
```

That seems to be the "Go" way, judging by their testing style. But I'm not tied to either idiom.

mattwomple:
> Just to confirm: This doesn't actually cause a response body to be emitted, right? I'm guessing from
> https://golang.org/src/net/http/server.go#L372 that this is the case, but I imagine you've tested.

Right.

> Or maybe make a loop like...

Thanks for the suggestion. I opted for looping over the array of anonymous structs. Let me know if you're ok with the style.

PR #291, opened 2019-04-30
signer: Avoid mutating response until signing

### LINKED ISSUE #279: Don't mutate fetch response until after all `proxy()` calls
Mutated response would ideally only be served in a SXG. i.e. move all the `Header.{Set,Del}` calls from inside the `case 200` to just after the `Set("Link")` call in `serveSignedExchange`. (Remove the `Del("Link")`.)

mattwomple:
After successfully (200) fetching an AMP document, complete all
validation steps that could result in proxying the original response
before mutating any headers.

Resolves #279

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -476,7 +476,8 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 		fetchResp.Header.Del(header)
 	}
 
-	// Set or remove Link header.
+	// Set Link header if formatting returned a valid value, otherwise, delete
+	// it to ensure there are no privacy-violating Link:rel=preload headers.
 	if linkHeader != "" {
 		fetchResp.Header.Set("Link", linkHeader)
 	} else {
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -348,17 +348,16 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
 		proxy(resp, fetchResp, nil)
 		return
 	}
+	var act string
 	var transformVersion int64
 	if this.requireHeaders {
 		header_value := GetJoined(req.Header, "AMP-Cache-Transform")
-		var act string
 		act, transformVersion = amp_cache_transform.ShouldSendSXG(header_value)
 		if act == "" {
 			log.Println("Not packaging because AMP-Cache-Transform request header is invalid:", header_value)
 			proxy(resp, fetchResp, nil)
 			return
 		}
-		resp.Header().Set("AMP-Cache-Transform", act)
 	} else {
 		var err error
 		transformVersion, err = transformer.SelectVersion(nil)
@@ -399,7 +398,7 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
 			return
 		}
 
-		this.serveSignedExchange(resp, fetchResp, signURL, transformVersion)
+		this.serveSignedExchange(resp, fetchResp, signURL, act, transformVersion)
 
 	case 304:
 		// If fetchURL returns a 304, then also return a 304 with appropriate headers.
@@ -442,7 +441,7 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 }
 
 // serveSignedExchange does the actual work of transforming, packaging and signed and writing to the response.
-func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *http.Response, signURL *url.URL, transformVersion int64) {
+func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *http.Response, signURL *url.URL, act string, transformVersion int64) {
 	// After this, fetchResp.Body is consumed, and attempts to read or proxy it will result in an empty body.
 	fetchBody, err := ioutil.ReadAll(io.LimitReader(fetchResp.Body, maxBodyLength))
 	if err != nil {
@@ -533,6 +532,13 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 	var body bytes.Buffer
 	if err := exchange.Write(&body); err != nil {
 		util.NewHTTPError(http.StatusInternalServerError, "Error serializing exchange: ", err).LogAndRespond(resp)
+	}
+
+	// If requireHeaders was true when constructing signer, the
+	// AMP-Cache-Transform outer response header is required (and has already
+	// been validated)
+	if (act != "") {
+		resp.Header().Set("AMP-Cache-Transform", act)
 	}
 
 	// TODO(twifkak): Add Cache-Control: public with expiry to match when we think the AMP Cache
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -387,16 +387,7 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request, param
 				proxy(resp, fetchResp, nil)
 				return
 			}
-			fetchResp.Header.Del(header)
 		}
-
-		// Mutate the fetched CSP to make sure it cannot break AMP pages.
-		fetchResp.Header.Set(
-			"Content-Security-Policy",
-			MutateFetchedContentSecurityPolicy(
-				fetchResp.Header.Get("Content-Security-Policy")))
-
-		fetchResp.Header.Del("Link") // Ensure there are no privacy-violating Link:rel=preload headers.
 
 		if fetchResp.Header.Get("Variants") != "" || fetchResp.Header.Get("Variant-Key") != "" ||
 			// Include versioned headers per https://github.com/WICG/webpackage/pull/406.
@@ -452,8 +443,6 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 
 // serveSignedExchange does the actual work of transforming, packaging and signed and writing to the response.
 func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *http.Response, signURL *url.URL, transformVersion int64) {
-	fetchResp.Header.Set("X-Content-Type-Options", "nosniff")
-
 	// After this, fetchResp.Body is consumed, and attempts to read or proxy it will result in an empty body.
 	fetchBody, err := ioutil.ReadAll(io.LimitReader(fetchResp.Body, maxBodyLength))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


twifkak:
Awesome! I haven't taken a look yet, but from the subject, I'm guessing this fixes #279?

mattwomple:
Results from testing `wompdbg.cmphys.com`:

`accept: text/html`

    server:nginx/1.14.0 (Ubuntu)
    date:Tue, 30 Apr 2019 03:30:52 GMT
    content-type:text/html
    content-length:1345
    last-modified:Tue, 09 Apr 2019 00:37:05 GMT
    connection:keep-alive
    etag:"5cabe931-541"
    strict-transport-security:max-age=31536000; includeSubdomains
    x-frame-options:SAMEORIGIN
    x-xss-protection:1; mode=block
    x-content-type-options:nosniff
    referrer-policy:no-referrer-when-downgrade
    vary:Accept, AMP-Cache-Transform
    accept-ranges:bytes

`accept: application/signed-exchange;v=b3` and `amp-cache-transform: google` with a dummy error at the location of comment _Begin mutations on original fetch response. From this point forward, do not fall-back to proxy()._

    server:nginx/1.14.0 (Ubuntu)
    date:Tue, 30 Apr 2019 03:11:06 GMT
    content-type:text/html
    content-length:1345
    connection:keep-alive
    accept-ranges:bytes
    amp-cache-transform:google;v="1"
    etag:"5cabe931-541"
    last-modified:Tue, 09 Apr 2019 00:37:05 GMT
    referrer-policy:no-referrer-when-downgrade
    strict-transport-security:max-age=31536000; includeSubdomains
    vary:Accept, AMP-Cache-Transform
    x-content-type-options:nosniff
    x-frame-options:SAMEORIGIN
    x-xss-protection:1; mode=block

`accept: application/signed-exchange;v=b3` and `amp-cache-transform: google` with a successful signing:

    server:nginx/1.14.0 (Ubuntu)
    date:Tue, 30 Apr 2019 03:21:08 GMT
    content-type:application/signed-exchange;v=b3
    transfer-encoding:chunked
    connection:keep-alive
    amp-cache-transform:google;v="1"
    cache-control:no-transform
    vary:Accept, AMP-Cache-Transform
    x-content-type-options:nosniff

mattwomple:
`-1` for now. I'm not certain this is ready. I wouldn't expect `amp-cache-transform` to be applied to the proxied response and I need to inspect the headers on the signed document.

mattwomple:
> https://github.com/ampproject/amppackager/blob/9cd3337d99521b2e07d1bb13a13b411028bee3e2/packager/signer/signer.go#L361
> 
> 
> I believe this should be moved until after any proxied response, but then, it could be useful for debugging proxied responses resulting from an error (informs end user what version of packaging is being used).
> Thoughts on this?

My inclination is to move this into `serveSignedExchange`, as I don't think it is necessary to add it to a 304 (not modified) redirect. Would you agree?

twifkak:
Sorry for the delay! Taking a look now. Yes, your inclination is the same as mine in both cases. [The amppkg documentation](https://github.com/ampproject/amppackager#productionizing) tells people to add `Vary` to all responses. And [the ACT spec](https://github.com/ampproject/amphtml/blob/master/spec/amp-cache-transform.md#response-header) is unclear, but I think it's safe to assume that by "if" it means "if and only if". Especially because I wrote it. :)

packager/signer/signer.go
```diff
@@ -470,17 +459,42 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
...
+	// Remove stateful headers.
+	for header := range statefulResponseHeaders {
+		fetchResp.Header.Del(header)
+	}
+
+	// Set or remove Link header.
 	if linkHeader != "" {
 		fetchResp.Header.Set("Link", linkHeader)
+	} else {
+		fetchResp.Header.Del("Link")
```
twifkak: Can you leave a comment in this case that deletion is necessary to meet the Google AMP Cache's requirements for privacy-preserving prefetch? Otherwise, I feel like this reason will get lost over time, and future amppkg devs might accidentally remove this line.

mattwomple: This should be resolved with cb93740

mattwomple:
> Thanks! This is excellent. Would you mind modifying one or more of the [ProxyUnsigned tests](https://github.com/ampproject/amppackager/blob/releases/packager/signer/signer_test.go#L405) to test this? Some/most cases is fine, if full coverage turns out to be difficult for some reason.

Thanks for the review. I'll be happy to add/update tests this weekend.

twifkak:
Thanks! Apologies in advance for the verbosity/complexity of the tests. Should be easy enough to follow the pattern, though.

mattwomple:
> Thanks! This is excellent. Would you mind modifying one or more of the [ProxyUnsigned tests](https://github.com/ampproject/amppackager/blob/releases/packager/signer/signer_test.go#L405) to test this? Some/most cases is fine, if full coverage turns out to be difficult for some reason.

Updated tests available. Don't be afraid to let me know if you're looking for something different.

I didn't figure out a good way to test content-length header since transformation would need to be performed.

twifkak:
> Edit: Following-up on what I said, "if you're looking for something different." I chose to generate a transformer error because it is close to the last type of error that can be encountered before signing occurs.

This makes sense. I don't think it's likely we'll have any other errors after that; only thing I can think of is if the signer starts validating something about the transformed AMP. But really I can't see why it would do that; requires re-parsing the HTML as opposed to `transformer.Process()` which already has a parse tree.

I would document in the test case that we're using transformer error because it's the last error before signing. That way, at the very least, if we *do* add another error, somebody may eventually notice the mismatch and fix it.

> I was tempted to move `if !this.shouldPackage()` to become the last use of `proxy()`, but if its definition in `cmd/gateway_server/server.go` ever becomes more than `return true`, then it's probably best not to delay short circuiting `ServeHTTP()`.

`gateway_server` is just for Google-internal integration tests. There's another use for shouldPackage, though: in production, it's false when `cert_cache` is unhealthy (i.e. doesn't have a valid OCSP response). It may do other things in the future (e.g. we'd planned to add a similar notion of health to the rtv cache). In any case, it should be a rare enough event that we needn't optimize it. However, I think the transformer error is probably the better solution; it seems like a natural last-error thing, because transormation is a costly thing we'd want to short-circuit, thus less likely to become not-the-last-error over time.

> Would it be appropriate to introduce yet another variable just for testing, like `testProxy`, which could have a comment by it to stipulate that it be the last use of `proxy()`? Or is that adding too many params to `ServeHTTP()`?

Can't easily add params to `ServeHTTP()`; it implements the [`httprouter.Handle`](https://godoc.org/github.com/julienschmidt/httprouter#Handle) type. ITYM `serveSignedExchange()` or `signer.New()`? Yeah, those already have too many params, so I'm hesitant to add more. If you can think of ways to clean that up, I'd be amenable. Maybe a future PR, though; no need to feature-creep this one.

packager/signer/signer_test.go
```diff
@@ -609,6 +622,33 @@ func (this *SignerSuite) TestProxyTransformError() {
 	this.Assert().Equal(fakeBody, body, "incorrect body: %#v", resp)
 }
 
+func (this *SignerSuite) TestProxyHeadersUnaltered() {
+	urlSets := []util.URLSet{{
+		Sign: &util.URLPattern{[]string{"https"}, "", this.httpsHost(), stringPtr("/amp/.*"), []string{}, stringPtr(""), false, 2000, nil},
+	}}
+
+	// Generate a request for non-existent transformer that will fail
```
twifkak: Document the reason here.

packager/signer/signer_test.go
```diff
@@ -609,6 +622,33 @@ func (this *SignerSuite) TestProxyTransformError() {
...
+	}
+	this.fakeHandler = func(resp http.ResponseWriter, req *http.Request) {
+		resp.Header().Set("Content-Type", "text/html")
+		resp.Header().Set("Set-Cookie", "chocolate chip")
+		resp.Header().Set("Cache-Control", "max-age=31536000")
+	}
+	resp := this.get(this.T(), this.new(urlSets), "/priv/doc?sign="+url.QueryEscape(this.httpsURL()+fakePath))
+	this.Assert().Equal(200, resp.StatusCode)
+
+	this.Assert().Equal("text/html", resp.Header.Get("content-type"))
```
twifkak: Is it possible to enumerate the full list here? e.g. [`this.Assert().ElementsMatch(...)`](https://godoc.org/github.com/stretchr/testify/assert#Assertions.ElementsMatch) or something. To prevent future changes from adding headers to the unsigned response.

twifkak:
(To be clear, I'm fine with adding the `act` param to `serveSignedExchange`. Seems necessary given the current structure of the code.)

packager/signer/signer_test.go
mattwomple: This should be resolved in 2e0abcc

packager/signer/signer_test.go
mattwomple: I took a stab at something like this in 2e0abcc . Let me know if it meets your expectations. Thanks again for the reviews!

twifkak:
Github seems to have posted my comments out of order. Start with https://github.com/ampproject/amppackager/pull/291#discussion_r281844572.

packager/signer/signer_test.go
```diff
@@ -627,26 +627,42 @@ func (this *SignerSuite) TestProxyHeadersUnaltered() {
...
 	resp := this.get(this.T(), this.new(urlSets), "/priv/doc?sign="+url.QueryEscape(this.httpsURL()+fakePath))
 	this.Assert().Equal(200, resp.StatusCode)
 
-	this.Assert().Equal("text/html", resp.Header.Get("content-type"))
-	this.Assert().Equal("chocolate chip", resp.Header.Get("set-cookie"))
-	this.Assert().Equal("max-age=31536000", resp.Header.Get("cache-control"))
-	this.Assert().Equal("", resp.Header.Get("content-security-policy"))
-	this.Assert().Equal("", resp.Header.Get("x-content-type-options"))
-	this.Assert().Equal("", resp.Header.Get("amp-cache-transform"))
+	for key, values := range resp.Header {
```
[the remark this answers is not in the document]
mattwomple: Your _reduced control flow_ routine has been adopted in commit 22c1608

packager/signer/signer_test.go
```diff
@@ -627,26 +627,42 @@ func (this *SignerSuite) TestProxyHeadersUnaltered() {
...
+		if key == "Vary" { // Vary is set for any request that reaches the signer
+			this.Assert().Equal(resp.Header.Get(key), "Accept, AMP-Cache-Transform")
+		} else if key == "Date" { // Allow Date to be updated
+			continue
+		} else { // Original headers must be unaltered and no new headers may appear
+			for _, value := range values {
+				this.Assert().Equal(value, originalHeaders[key])
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mattwomple: This should be addressed in commit 22c1608

packager/signer/signer_test.go
```diff
@@ -627,26 +627,42 @@ func (this *SignerSuite) TestProxyHeadersUnaltered() {
...
-	this.Assert().Equal("", resp.Header.Get("content-security-policy"))
-	this.Assert().Equal("", resp.Header.Get("x-content-type-options"))
-	this.Assert().Equal("", resp.Header.Get("amp-cache-transform"))
+	for key, values := range resp.Header {
+		if key == "Vary" { // Vary is set for any request that reaches the signer
+			this.Assert().Equal(resp.Header.Get(key), "Accept, AMP-Cache-Transform")
+		} else if key == "Date" { // Allow Date to be updated
+			continue
+		} else { // Original headers must be unaltered and no new headers may appear
+			for _, value := range values {
```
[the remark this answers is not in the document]
mattwomple: This should be addressed in commit 22c1608

mattwomple:
Thanks again for the great review.

twifkak:
Thanks! I'm out of nits to pick, so I will approve and merge.

Issue #293, opened 2019-05-03
Add more tests to accept_test.go


twifkak:
Add all the tests in cl/243547227 that are missing from [`accept_test.go`](https://github.com/ampproject/amppackager/blob/releases/packager/accept/accept_test.go). Maybe require putting some of those on the wrong side, as #240 is still open.

twifkak:
Also, per https://github.com/ampproject/amppackager/pull/296#discussion_r285801253:

> Add some tests including `\\"` (escaped backslash followed by unescaped quote) and `\\\"` (escaped backslash followed by escaped quote), to guard against `tokenize` being replaced by something simpler but wrong.

twifkak:
And maybe https://github.com/ampproject/amppackager/pull/296#discussion_r285801652, if a falsifiable test can be created:

> I think if `!inQuotes`, this should be a parse error, per the [tchar definition](https://tools.ietf.org/html/rfc7230#section-3.2.6). I guess returning empty list would make sense in this case. Add a test for this if possible; maybe something like `application/signed-exchange;v=b3;x=foo\,bar`.

PR #296, opened 2019-05-06
accept: Improve tokenization

### LINKED ISSUE #240: Fix Accept parsing
There is a bug in the comma-splitting logic for Accept (not fixed in #196). It doesn't ignore commas inside quoted strings. A correct splitter would need to identify `,`, `"`, and `\"`. See [Spring's tokenizer](https://www.google.com/url?q=https://github.com/spring-projects/spring-framework/blob/7bbd4c681fd009898a62cac6e11eae77039a4119/spring-core/src/main/java/org/springframework/util/MimeTypeUtils.java%23L274&sa=D&source=hangouts&ust=1548433024564000&usg=AFQjCNFkqlocOoRDdAL9J8mObQTnU_fTBw) for example.

This bug is mentioned in `accept_test.go`, but downplayed as rare. However, it is likely to occur in the future when Chrome advertises support for `v="b3,b4"`.

mattwomple:
- Support parameter values with commas (inside quotes)
- Add a few relevant tests for new support

Tokenization method referenced from spring-projects, which uses the
compatible Apache 2.0 license:
  https://github.com/spring-projects/spring-framework/

Fixes #240


twifkak:
@alin04 In case you have time to review.

packager/accept/accept.go
```diff
@@ -18,23 +18,57 @@ const SxgContentType = "application/signed-exchange;v=" + AcceptedSxgVersion
...
+	var tokens []string
+	acceptLen := len(accept)
+	if acceptLen == 0 {
+		return tokens
+	}
+
+	inQuotes := false
+	startIndex := 0
+	for i := 0; i < acceptLen; i++ {
+		char := string(accept[i])
```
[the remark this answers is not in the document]
mattwomple: Thanks for the suggestion, changed in commit fe1d37fabe8c62300389b08867a46b4717c42e4d

packager/accept/accept.go
```diff
@@ -18,23 +18,57 @@ const SxgContentType = "application/signed-exchange;v=" + AcceptedSxgVersion
...
+	inQuotes := false
+	startIndex := 0
+	for i := 0; i < acceptLen; i++ {
+		char := accept[i]
+		switch char {
+		case '"':
+			inQuotes = !inQuotes;
+		case ',':
+			if !inQuotes {
+				tokens = append(tokens, strings.TrimSpace(accept[startIndex:i]))
```
twifkak: [TrimSpace](https://golang.org/pkg/strings/#TrimSpace) trims too much. Use [TrimFunc](https://golang.org/pkg/strings/#TrimFunc) with a function that just checks for `' '` or `'\t'` (per the [#rule definition](https://tools.ietf.org/html/rfc7230#section-7)). Maybe define that function in [packager/util/http.go](https://github.com/ampproject/amppackager/blob/master/packager/util/http.go).

Ditto on line 45.

(Maybe this isn't a problem in practice? I'd prefer to be precise anyway.)

packager/accept/accept_test.go
```diff
@@ -11,16 +11,17 @@ func TestCanSatisfy(t *testing.T) {
 	assert.False(t, CanSatisfy("*/*"))
 	assert.False(t, CanSatisfy("image/jpeg;v=b3"))
 	assert.False(t, CanSatisfy(`application/signed-exchange;v=b2`))
-	// This is a bug that will be triggered when a UA starts supporting multiple SXG versions:
-	assert.False(t, CanSatisfy(`application/signed-exchange;x="a,b";v="b3"`))
+	assert.False(t, CanSatisfy(`application/signed-exchange;v="b1,b2"`))
+	assert.False(t, CanSatisfy(`application/signed-exchange;x="y,application/signed-exchange;v=b3,z";v=b1`))
```
twifkak: Add some tests including `\\"` (escaped backslash followed by unescaped quote) and `\\\"` (escaped backslash followed by escaped quote), to guard against `tokenize` being replaced by something simpler but wrong.

packager/accept/accept.go
```diff
@@ -18,23 +18,57 @@ const SxgContentType = "application/signed-exchange;v=" + AcceptedSxgVersion
...
+				startIndex = i + 1;
+			}
+		case '\\':
+			i++
```
twifkak: I think if `!inQuotes`, this should be a parse error, per the [tchar definition](https://tools.ietf.org/html/rfc7230#section-3.2.6). I guess returning empty list would make sense in this case. Add a test for this if possible; maybe something like `application/signed-exchange;v=b3;x=foo\,bar`.

(Perhaps `mime.ParseMediaType` already catches this? Ditto above; I'd prefer precision / defense in depth.)

packager/accept/accept.go
```diff
@@ -18,23 +18,57 @@ const SxgContentType = "application/signed-exchange;v=" + AcceptedSxgVersion
...
+		return tokens
+	}
+
+	inQuotes := false
+	startIndex := 0
+	for i := 0; i < acceptLen; i++ {
+		char := accept[i]
+		switch char {
+		case '"':
+			inQuotes = !inQuotes;
```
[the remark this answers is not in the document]
mattwomple: This is addressed in 00a004056af7d05c92365a950948d5934dfe6201

packager/accept/accept.go
mattwomple: This is addressed in a457c3273479f11f489be8ca02a07cb589eb3041

packager/accept/accept_test.go
mattwomple: Could issue #293 be used for this request?

packager/accept/accept.go
mattwomple: This is addressed in b974b06fab6008e2e0c173060108eed6b4e2ec7c

mattwomple:
I'm sorry this has taken so long. Thank you for the careful review. My free time has vanished and won't return for another few weeks. The option "Allow edits from maintainers." is enabled, so please feel free to push additional changes as you see fit. I hope to keep playing with `amppackager` in the future, and I appreciate what I've learned, largely thanks to your reviews.

packager/accept/accept_test.go
twifkak: It's not addressed by the CL in question, but I'm okay moving this into that issue.

PR #299, opened 2019-05-07
Remove warning about code, and add one about spec.

twifkak:
Remove the warning about the code being in developer preview; we
consider it stable enough to deploy in production now. Both code and
specification are still evolving, but this is handled by the requirement
to keep the packager updated per the `AMP-Cache-Transform` version.

Add a warning about SXG security considerations. This is an attempt to
distill advice from the risks outlined in the spec.

cc @jyasskin

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,5 @@
 # AMP Packager
 
-> **WARNING**: This code is still evolving, and is a developer preview. The
-> specification is still changing, and this is an implementation of a snapshot
-> of it. Feel free to use it, but treat with care how you configure and deploy it.
-
 AMP Packager is a tool to [improve AMP
 URLs](https://www.ampproject.org/latest/blog/developer-preview-of-better-amp-urls-in-google-search).
 By running it in a proper configuration, web publishers may (eventually) have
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -150,6 +146,21 @@ packager URL directly, first add a Chrome extension to send an
...
+Signed exchanges come with some [security
+considerations](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#security-considerations)
+that publishers should consider. A starting list of recommendations based on
+that:
+
+ * Use different keys for the signed exchange cert and the TLS cert.
+ * Only sign public content that's OK to be shared with crawlers.
+ * Don't sign personalized content. (It's OK to sign content that includes
+   static JS that adds personalization at runtime.)
+ * Be careful when signing inline JS; if it includes a vulnerability, it may be
```
twifkak: Vulnerabilities exist outside of JS, but I was trying to provide slightly more concrete advice than the spec, based on my belief of where they're the most prevalent. Perhaps there's an option that's both general and concrete? Or perhaps just stick with abstract in this case?

Gregable: I think it's OK. The part I don't like is "Be careful" isn't very actionable, not that you are trying to have a specific opinion beyond the discussion in the draft.

PR #304, opened 2019-05-09
Document the new warning header.

twifkak:
The cache recently started outputting this to help with SXG errors.


twifkak:
Yeah, that change SGTM. Once it lands in production, we can update the docs here.

PR #306, opened 2019-05-10
Check certificate profiles and its validity period

shigeki:
This makes the following checks to the leaf certificate if it has
- a valid public key corresponding to its private key
- a domain in CN or SubjectAltname specified in URLSet.Sign.Domain
- a validity period within 90 days according to SXG cert requirements


shigeki:
After 2019-08-01, we should cleanup this fix and `testdata/b1/*`.

twifkak:
Nice! Despite not having a bug #, this seems pretty important for usability. I'm on vacation for the next week, so I may not have time to review until I get back. Hope that's okay.

@alin04 In case you have time to review.

cmd/amppkg/main.go
```diff
@@ -94,13 +94,19 @@ func main() {
...
 
 	key, err := util.ParsePrivateKey(keyPem)
 	if err != nil {
 		die(errors.Wrapf(err, "parsing %s", config.KeyFile))
 	}
-	// TODO(twifkak): Verify that key matches certs[0].
+
+	for i := range config.URLSet {
+		domain := config.URLSet[i].Sign.Domain
+		err := util.CheckCertificate(certs[0], key, domain)
```
twifkak: Optionally:

```
if err := util.CheckCertificate(certs[0], key, domain); err != nil {
  die(...)
}
```

packager/util/util.go
```diff
@@ -76,3 +78,32 @@ func CanSignHttpExchanges(cert *x509.Certificate) bool {
...
+	if certPubKey.Y.Cmp(pubKey.Y) != 0 {
+		return errors.New("PublicKey.Y not match")
+	}
+	if err := cert.VerifyHostname(domain); err != nil {
+		return err
+	}
+	// https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#cross-origin-cert-req
+	// Clients MUST reject certificates with this extension that were issued after 2019-05-01 and have a Validity Period longer than 90 days.
+	// After 2019-08-01, clients MUST reject all certificates with this extension that have a Validity Period longer than 90 days.
+	may1st := time.Date(2019, time.May, 1,  0, 0, 0, 0, time.UTC)
```
twifkak: Move these to the top-level outside the function, and give them semantic names, e.g.:

```
// https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#cross-origin-cert-req
// Clients MUST reject certificates with this extension that were issued after 2019-05-01 and have a Validity Period longer than 90 days.
// After 2019-08-01, clients MUST reject all certificates with this extension that have a Validity Period longer than 90 days.
var start90DayGracePeriod = time.Date(2019, time.May, 1,  0, 0, 0, 0, time.UTC)
var end90DayGracePeriod = time.Date(2009, time.August, 1, 0, 0, 0, 0, time.UTC)
```

packager/util/util_test.go
```diff
@@ -27,3 +35,41 @@ func TestCanSignHttpExchanges(t *testing.T) {
...
+}
+
+func TestParse91DaysCertificate(t *testing.T) {
+	assert.Contains(t, errorFrom(util.CheckCertificate(pkgt.B3Certs91Days[0],
+		pkgt.B3Key, "amppackageexample.com")), "Certificate MUST have a Validity Period no greater than 90 days")
+}
+
+func TestParseCertificateIssuedBeforeMay1(t *testing.T) {
+	aug1st := time.Date(2009, time.August, 1, 0, 0, 0, 0, time.UTC)
+	if time.Now().After(aug1st) {
```
twifkak: Is it possible to make this test stable in the face of an ever-changing `time.Now()`? I suggest one of two options:

 - Pass `now time.Date` as another param to `util.CheckCertificate`.
 - Temporarily modify `start90DayGracePeriod` and `end90DayGracePeriod` (introduced per my other suggestion), to elicit the desired test conditions.

The benefit is that both cases are tested now -- we don't have a surprise test failure on 1 Aug. (I trust that you tested the Aug 1 case, but don't trust that future changes will.)

I have a slight preference for the first option, seems simpler, but not a strong preference.

testdata/b3/ca.cert
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
twifkak: Can you add a README.md to this directory, with instructions (e.g. shell commands) to generate these?

cmd/amppkg/main.go
```diff
@@ -94,13 +94,19 @@ func main() {
...
 	}
-	// TODO(twifkak): Verify that certs[0] covers all the signing domains in the config.
 
 	key, err := util.ParsePrivateKey(keyPem)
 	if err != nil {
 		die(errors.Wrapf(err, "parsing %s", config.KeyFile))
 	}
-	// TODO(twifkak): Verify that key matches certs[0].
+
+	for i := range config.URLSet {
```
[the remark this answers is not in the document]
shigeki: fixed

cmd/amppkg/main.go
shigeki: fixed

packager/util/util.go
shigeki: fixed

packager/util/util_test.go
shigeki: fixed with the first option.

testdata/b3/ca.cert
shigeki: Added

shigeki:
Thanks for reviewing. Fixed with your comments. PTAL.

PR #311, opened 2019-05-23
fix entries of hop-by-hop header

shigeki:
Remove unnecessary headers as

- Proxy-Authenticate: duplicated
- Proxy-Authorization: request header not response
- TE: not a hop-by-hop header

This can check the same response headers as Chrome does in
https://chromium.googlesource.com/chromium/src.git/+/master/content/browser/web_package/signed_exchange_envelope.cc#29


packager/util/http.go
```diff
@@ -35,10 +35,7 @@ var ConditionalRequestHeaders = map[string]bool{
 var legacyHeaders = map[string]bool{
 	"Connection": true,
 	"Keep-Alive": true,
-	"Proxy-Authenticate": true,
-	"Proxy-Authorization": true,
```
twifkak: Good catch, thanks!

packager/util/http.go
```diff
@@ -35,10 +35,7 @@ var ConditionalRequestHeaders = map[string]bool{
...
 	"Proxy-Connection": true,
-	"TE": true,
```
twifkak: Good catch, thanks! But I think this is for the same reason as `Proxy-Authorization`, no? https://tools.ietf.org/html/rfc2616#section-13.5.1 say it's a hop-by-hop header, but https://tools.ietf.org/html/rfc2616#section-14.39 says it's a request header. Did I miss something?

packager/util/http.go
```diff
@@ -35,10 +35,7 @@ var ConditionalRequestHeaders = map[string]bool{
 var legacyHeaders = map[string]bool{
 	"Connection": true,
 	"Keep-Alive": true,
-	"Proxy-Authenticate": true,
```
twifkak: I assume the duplication you're referring to is `statefulResponseHeaders` in `signer.go`. I think I prefer that duplication, because it's semantically attached to a different concept (RFC2616 vs SXG spec). For instance, if in the future, SXG v=b4 no longer disallows `Proxy-Authenticate`, we still want to remove it here because we're acting as a gateway. Or perhaps in the future, we have a pluggable fetch API; we still want it removed by `signer.go` so it doesn't generate an invalid SXG.

packager/util/http.go
shigeki: I also referred the spec of 
https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#rfc.section.4.1 for a hop-by-hop header list but rfc2616 you cited and  https://tools.ietf.org/html/rfc7230#section-4.3 says that

> Since the TE header field only applies to the immediate connection, a sender of TE MUST also send a "TE" connection option within the Connection header field (Section 6.1) in order to prevent the TE field from being forwarded by intermediaries that do not support its semantics.

It means we should include `TE` in a hop-by-hop as an intermediaries.  I will fix this.

packager/util/http.go
shigeki: Yes, I will fix this.

packager/util/http.go
twifkak: I think that means that if a client includes a `TE` request header, it should also include a `Connection: TE` request header, for servers that receive that request. I don't think that applies to the response path.

twifkak: (To be clear, I think there's nothing to fix -- I think your original proposed change is correct. But I may be wrong; I have no experience with TE.)

shigeki:
Fixed and force-pushed to change this commit message.

packager/util/http.go
shigeki: Oops, that request header not response! I will fix this again.

shigeki:
Fixed. PTAL.

PR #312, opened 2019-05-23
Check not to have TE in config.ForwardedRequestHeaders

shigeki:
Per #311, I missed that `TE` must not be included in `config.ForwardedRequestHeaders`. 

For `Proxy-Authorization`,  it can be forwarded as
> A proxy MAY relay the credentials from the client request to the next proxy if that is the mechanism by which the proxies cooperatively authenticate a given request.

described in https://tools.ietf.org/html/rfc7235#section-4.4.

### DIFF
--- a/packager/util/config_test.go
+++ b/packager/util/config_test.go
@@ -115,6 +115,18 @@ func TestForwardedRequestHeadersHaveDisallowedHeader(t *testing.T) {
 	`))), "ForwardedRequestHeaders must not include request header of via")
 }
 
+func TestForwardedRequestHeadersHaveTE(t *testing.T) {
+	assert.Contains(t, errorFrom(ReadConfig([]byte(`
+		CertFile = "cert.pem"
+		KeyFile = "key.pem"
+		OCSPCache = "/tmp/ocsp"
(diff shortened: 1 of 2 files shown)


twifkak:
Good catch. Re: `Proxy-Authorization`, though, I think the SXG [section on over-signing](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#seccons-over-signing) is relevant:

> Signing systems SHOULD also incorporate the following mitigations to reduce the risk that private responses are signed:
> 1. Strip the Cookie request header field and other identifying information like client authentication and TLS session IDs from requests whose exchange is destined to be signed, before forwarding the request to a backend.

That said, I think it would be a highly unusual setup for a user-agent to be configured to a (forward) proxy that then forwards that request to `amppkg` with `proxy-authorization` header. So I'm not _too_ worried about it.

I'd suggest either also remove it here, or just add a short warning in `amppkg.example.toml` (given how unlikely it would be accidentally misused).

shigeki:
> I'd suggest either also remove it here,

I agree. Proxy-Authorization should not be included in config.ForwardedRequestHeaders.
Fixed.

PR #313, opened 2019-05-24
Switch to hand-rolled http mux.

### LINKED ISSUE #310: Ensure inner URL = outer URL where possible
```
curl -s -H 'Accept: application/signed-exchange;v=b3' -H 'AMP-Cache-Transform: any' \
  https://amppackageexample.com/inde%78.html
```

returns an SXG whose inner URL has had the %78 unescaped. This doesn't meet Google's [cache requirements](https://github.com/ampproject/amppackager/blob/releases/docs/cache_requirements.md).

amppkg should try to ensure inner URL = outer URL as much as possible (i.e. where doing so doesn't contradict the SXG [requirements for `fallbackUrl`](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange) or the [recommendations to mitigate content sniffing](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#seccons-content-sniffing)).

twifkak:
This fixes the error where `/priv/doc/https://` URLs were being
improperly unescaped before fetching/signing.

In addition, validate that sign URLs don't contain any special characters, per the WICG recommendation.

The new mux code depends directly on the handlers, because we don't need any fancy dependency inversion. This required movings tests into a different package to break the import cycle, which in turn required changing the visibility of a lot of things.

Resolves #310.


packager/mux/mux.go
```diff
@@ -0,0 +1,93 @@
+package mux
```
alin04: Missing copyright.

packager/mux/mux.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ampproject/amppackager/packager/validitymap"
+)
+
+type mux struct {
+	certCache *certcache.CertCache
+	signer *signer.Signer
+	validityMap *validitymap.ValidityMap
+}
+
+// Implements the HTTP routing strategy used by the amppkg server. The main
```
alin04: Maybe move all/most of these comments to package level.

packager/mux/mux.go
```diff
@@ -0,0 +1,93 @@
...
+func tryTrimPrefix(s, prefix string) (string, bool) {
+	sLen := len(s)
+	trimmed := strings.TrimPrefix(s, prefix)
+	return trimmed, len(trimmed) != sLen
+}
+
+var allowedMethods = map[string]bool{http.MethodGet: true, http.MethodHead: true}
+
+func (this *mux) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	if _, ok := allowedMethods[req.Method]; !ok {
```
alin04: This only checks for key existence, and not the value. What if the value was false in the map?

packager/mux/mux.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Use EscapedPath rather than RequestURI because the latter can take
+	// absolute-form, per https://tools.ietf.org/html/rfc7230#section-5.3.
+	//
+	// Use EscapedPath rather than RawPath to verify that the request URI
+	// is a valid encoding, per
+	// https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange
+	// item 3.
+	path := req.URL.EscapedPath()
+	if suffix, ok := tryTrimPrefix(path, "/priv/doc"); ok {
```
alin04: Move "/priv/doc" into a global const var somewhere and dedupe it with main.go where it's also referenced. Or pass in "/priv/doc" as a parameter.

packager/mux/mux.go
```diff
@@ -0,0 +1,93 @@
...
+				params["signURL"] += "?" + req.URL.RawQuery
+			}
+			this.signer.ServeHTTP(resp, req)
+		} else {
+			http.NotFound(resp, req)
+		}
+	} else if suffix, ok := tryTrimPrefix(path, util.CertURLPrefix + "/"); ok {
+		unescaped, err := url.PathUnescape(suffix)
+		if err != nil {
+			http.Error(resp, "400 bad request", http.StatusBadRequest)
```
alin04: More descriptive error message than "400 bad request"?

packager/mux/params/params.go
```diff
@@ -0,0 +1,32 @@
+// A separate module to avoid import cycles between mux and the various
```
alin04: copyright

packager/signer/signer.go
```diff
@@ -265,19 +267,20 @@ func (this *Signer) genCertURL(cert *x509.Certificate, signURL *url.URL) (*url.U
...
 	}
 	var fetch, sign string
-	if inPathSignURL := params.ByName("signURL"); inPathSignURL != "" {
-		sign = inPathSignURL[1:] // Strip leading "/" produced by httprouter.
-		if req.URL.RawQuery != "" {
-			sign += "?" + req.URL.RawQuery
-		}
+	params := muxp.Params(req)
+	if inPathSignURL := params["signURL"]; inPathSignURL != "" {
+		println("inPathSignURL =", inPathSignURL)
```
alin04: log.Println ?

packager/mux/mux.go
twifkak: Done.

packager/mux/mux.go
twifkak: Done.

packager/mux/mux.go
twifkak: Done.

packager/mux/mux.go
twifkak: It's not mentioned in main.go; this is the only production usage.

It's in signer_test.go, and I followed that same pattern with "/amppkg/cert" (certcache_test.go, signer_test.go) and "/amppkg/validity" (validitymap_test.go, signer_test.go). Vaguely thinking that it's better for tests to be explicit. I could change it if you think I should, but I'd do it for all, to be consistent.

packager/mux/params/params.go
twifkak: Merged into mux.go.

packager/mux/mux.go
twifkak: Done.

packager/signer/signer.go
twifkak: Oops, removed all of these.

twifkak:
Test for the escaping is in signer_test.
 - For the validation.go change, see `TestEscapeQueryParamsInFetchAndSign` and `TestDisallowInvalidCharsSign`. (Hmm, perhaps that should be in validation_test.go.)
 - For the mux change, see the 3 `TestSignAsPathParam*` tests.

twifkak:
Also, got rid of all the capitalization changes & the need for the separate `params` pkg.

packager/mux/mux.go
alin04: https://github.com/ampproject/amppackager/blob/releases/cmd/amppkg/main.go#L139 ?

twifkak: Removed in this PR.

packager/signer/validation.go
```diff
@@ -123,10 +123,34 @@ func fetchURLMatches(url *url.URL, pattern *util.URLPattern) error {
...
+	// Vaguely ordered most to least common, to aid short-circuiting:
+	return (b >= 'a' && b <= 'z') || b == '_' || b == '~' ||
+		(b >= '!' && b < 'Z' && b != '"' /*x22*/ && b != '#' /*x23*/ && b != '<' /*x3C*/ && b != '>' /*x3E*/)
+}
+
 // True iff url matches pattern, as defined by an [URLSet.Sign] block in the
 // config file. The format of this URLPattern is validated by
 // validateSignURLPattern in config.go.
 func signURLMatches(url *url.URL, pattern *util.URLPattern) error {
+	for _, b := range []byte(url.String()) {
```
alin04: Optional, leverage Go's IndexFunc, e.g. if IndexFun(url.String(), isFallbackURLCodePoint) == -1 { ... }

twifkak: Could work? https://golang.org/pkg/bytes/#IndexFunc is based on a UTF-8 decoding of the string, whereas I want to go byte-by-byte. I think any multibyte UTF-8 char is also false for `isFallbackURLCodePoint`? But I don't want to have to think about it. :)

PR #316, opened 2019-06-07
Allow incorrect cert in development.

twifkak:
When `-development` is passed, an invalid SXG cert will only cause a
WARNING.

@shigeki

### DIFF
--- a/cmd/amppkg/main.go
+++ b/cmd/amppkg/main.go
@@ -90,8 +90,12 @@ func main() {
 	if certs == nil || len(certs) == 0 {
 		die(fmt.Sprintf("no cert found in %s", config.CertFile))
 	}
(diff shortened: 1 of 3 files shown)


shigeki:
Thanks, I missed it.

If `-development` is intended to work with `--ignore-certificate-errors-spki-list` on Chrome, only cert's validity period checks need to be warned. Other checks of cert's key pair and signing domain in `util.CheckCertificate` should be failed.

I think it is better to  move validity period checks out of `util.CheckCertificate` and allow it if `-developement` such as `util.CanSignHttpExchanges`.

twifkak:
@shigeki Thanks for the review.

It's not documented in README.md, but I also use `-development` with `curl ... | dump-signedexchange -verify`.

I moved the validity check from `CheckCertificate` to `CanSignHttpExchanges` (and changed its doc and interface accordingly). I also added support for the `-invalidcert` flag which I forgot. Does this seem reasonable?

PR #329, opened 2019-07-12
amppkg_dl_sxg: add -cert_url_path to override cert-url

shigeki:
During renewing of SXG cert in cluster servers, the new cert cannot
be obtained in requesting cert-url from the cluster. `-cert_url_path`
option enables us to overriding scheme, hostname and path of cert-url
so that we can obtain the newer cert by specifying a server.

In the server, we can get a new cert and confirm its cert renewal by
using localhost such as

`amppkg_dl_sxg -cert_url_path http://localhost:8080/amppkg/cert/ http://localhost:8080/priv/doc?...`

### DIFF
--- a/cmd/amppkg_dl_sxg/main.go
+++ b/cmd/amppkg_dl_sxg/main.go
@@ -7,7 +7,9 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
+	"path"
 	"regexp"
 	"strconv"
 
@@ -17,6 +19,7 @@ import (
 
 var flagOutSXG = flag.String("out_sxg", "test.sxg", "Path to where the signed-exchange should be saved.")
 var flagOutCert = flag.String("out_cert", "test.cert", "Path to where the cert-chain+cbor should be saved.")
+var flagCertUrlBase = flag.String("cert_url_base", "", "Override scheme, hostname and parent path in cert-url.")
 
 func getSXG(url string) ([]byte, error) {
 	req, err := http.NewRequest("GET", url, nil)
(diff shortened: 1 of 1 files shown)


cmd/amppkg_dl_sxg/main.go
```diff
@@ -17,6 +19,7 @@ import (
 
 var flagOutSXG = flag.String("out_sxg", "test.sxg", "Path to where the signed-exchange should be saved.")
 var flagOutCert = flag.String("out_cert", "test.cert", "Path to where the cert-chain+cbor should be saved.")
+var flagCertURL = flag.String("cert_url_path", "", "Override scheme, hostname and path in cert-url.")
```
twifkak: Rename to cert_url_base. (Maybe s/path/parent path/ in the flag description?)

cmd/amppkg_dl_sxg/main.go
```diff
@@ -90,11 +99,25 @@ func main() {
 	if err != nil {
 		log.Fatalf("%+v", err)
 	}
+	cURL, err := url.Parse(certURL)
+	if err != nil {
+		log.Fatalf("%+v", err)
+	}
+	fURL, err := url.Parse(*flagCertURL)
```
twifkak: Maybe move this url.Parse call inside the `if *flagCertURL`?

cmd/amppkg_dl_sxg/main.go
```diff
@@ -90,11 +99,25 @@ func main() {
...
+	if err != nil {
+		log.Fatalf("%+v", err)
+	}
+	if *flagCertURL != "" {
+		cURL.Scheme = fURL.Scheme
```
twifkak: How about something like:

```go
certHash := path.Base(cURL.Path)
cURL = fURL
cURL.Path = path.Join(cURL.Path, certHash)
```

That way, the user can provide a `user:pass@` or `?query` too. (Admittedly, rare cases.)

cmd/amppkg_dl_sxg/main.go
shigeki: Fixed.

cmd/amppkg_dl_sxg/main.go
shigeki: Fixed.

cmd/amppkg_dl_sxg/main.go
shigeki: Fixed.

shigeki:
Thanks for your review. Fixes were done according to your comments.

twifkak:
Thanks, merging.

PR #332, opened 2019-08-05
Add link to amp.dev docs for Signed Exchanges in AMP Packager Readme.

### LINKED ISSUE #305: Add link to amp.dev docs
README.md should link to https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/signed-exchange somewhere.

Gregable:
Fixes #305.

Please verify when reviewing that `master`, rather than `releases` is the correct branch to pull this into.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # AMP Packager
 
 AMP Packager is a tool to [Serve AMP using Signed Exchanges](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/signed-exchange/) which in turn [improves AMP
-URLs](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/signed-exchange/).
+URLs](https://blog.amp.dev/2018/11/13/developer-preview-of-better-amp-urls-in-google-search/).
 By running it in a proper configuration, web publishers may (eventually) have
 origin URLs appear in AMP search results.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,15 @@
 # AMP Packager
 
-AMP Packager is a tool to [improve AMP
-URLs](https://www.ampproject.org/latest/blog/developer-preview-of-better-amp-urls-in-google-search).
+AMP Packager is a tool to [Serve AMP using Signed Exchanges](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/signed-exchange/) which in turn [improves AMP
+URLs](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/signed-exchange/).
```
twifkak: I like leading with the purpose, the intended effect of fixing AMP URLs, and following with the means (signed exchanges). I'd also lean towards not having two links to the same URL in the same sentence; the user will probably click both not realizing they're the same (or at least hover, costing time). So, e.g.

```
AMP Packager is a tool to [improve AMP URLs](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/signed-exchange/)
by serving AMP using signed exchanges.
```

FWIW the second para goes into detail on signed exchanges, so the mention might not be necessary in the first para, but I'm okay with it either way.

Gregable: Oops. I meant to change the 2nd link to the correct blog post, rather than have them both be the same. If you'd prefer, I can just remove it.

Issue #93, opened 2018-09-13
Support short-term cert expiry


twifkak:
Currently, `amppkg` only loads the cert file at startup. If it expires while the packager is running, the packager continues to sign with it and serve it. Instead, it should attempt to reload automatically starting a few days before expiry, and continuing at some regular interval until no longer imminently expiring. If the cert is expired, it should stop signing exchanges, and log a warning.

In addition, it should serve the cert-url with an http expiry no longer than the cert expiry (as a follow-up to #85).

twifkak:
This will be much more important as of WICG/webpackage#383.

twifkak:
Restarting the server every 90 days is an option. Perhaps not a particularly pleasant one, but perhaps not so bad in this world of coordinated containers.

twifkak:
I believe cert renewals will have a different cert-sha256 (confirm this) and hence a different cert-url and generate different signatures. If so, we should decide whether to continue serving the old cert at the old URL (up until expiry). We may get a timeline like:

 1. AMP cache requests SXG, amppkg provides.
 2. amppkg has new cert.
 3. AMP cache requests cert-url, amppkg 404s.

There may be an arbitrary amount of time between (1) and (3), though likely usually small. Though the AMP cache is free to respond to the 404 by doing a GOTO 1.

I'm leaning to 'no' for simplicity, since AMP Caches have some workarounds available.

Alternatively, maybe we should reconsider data: cert-urls.

twifkak:
Update:

 - We should support hosting old and new cert at same time. (Does this mean we'll need to change the toml to allow a list of cert paths?)
 - Optional: Allow adding new cert without restarting. (Either inotify or SIGUSR1.)
 - Required, but later is ok: Integrate an ACME client library.

twifkak:
Sorry for the noise; changing things back since I see from the comments there is other stuff than just the tests. Split the tests off into #433 since that seems fixit-sized.

PR #338, opened 2019-08-20
Packager config mods

banaag:
Changes to amppackager toml file to accommodate ACME config changes.

Addresses #93

### DIFF
--- a/packager/util/config.go
+++ b/packager/util/config.go
@@ -24,13 +24,17 @@ import (
 )
 
 type Config struct {
-	LocalOnly bool
-	Port      int
-	CertFile  string // This must be the full certificate chain.
-	KeyFile   string // Just for the first cert, obviously.
-	OCSPCache string
+	LocalOnly	bool
+	Port		int
+	CertFile	string // This must be the full certificate chain.
+	KeyFile		string // Just for the first cert, obviously.
+	NewCertFile	string // The new full certificate chain replacing the expired one.
+	NewKeyFile	string // For the first cert in NewCertFile.
+	AutoRenewCert	bool   // Should we auto-renew cert? Defaults to false.
+	OCSPCache	string
 	ForwardedRequestHeaders []string
-	URLSet    []URLSet
+	URLSet		[]URLSet
+	ACMEConfig	*ACMEConfig
 }
 
 type URLSet struct {
@@ -48,6 +52,16 @@ type URLPattern struct {
 	ErrorOnStatefulHeaders bool
 	MaxLength              int
 	SamePath               *bool
+}
+
+type ACMEConfig struct {
+	Prod	*ACMEServerConfig
+	Staging	*ACMEServerConfig
+}
+
+type ACMEServerConfig struct {
+	DiscoURL	string // ACME Production Directory Resource URL
+	AccountURL	string // ACME Account URL
 }
 
 // TODO(twifkak): Extract default values into a function separate from the one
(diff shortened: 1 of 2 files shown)


twifkak:
@googlebot

packager/util/config.go
```diff
@@ -24,13 +24,17 @@ import (
...
-	LocalOnly bool
-	Port      int
-	CertFile  string // This must be the full certificate chain.
-	KeyFile   string // Just for the first cert, obviously.
-	OCSPCache string
+	LocalOnly	bool
+	Port		int
+	CertFile	string // This must be the full certificate chain.
+	KeyFile		string // Just for the first cert, obviously.
+	NewCertFile	string // The new full certificate chain replacing the expired one.
```
twifkak: Can you explain the semantics here? Currently, CertFile is only read by packager. Is NewCertFile read/write? What happens in 180-days with NewNewCertFile? Or is CertFile now read/write, with some process moving NewCertFile to CertFile before it's time for NewNew?

packager/util/config.go
```diff
@@ -24,13 +24,17 @@ import (
...
+	NewKeyFile	string // For the first cert in NewCertFile.
```
twifkak: I don't think NewKeyFile is necessary for v1. (Correct me if I'm wrong!) It looks like ACME does have a key change protocol (https://tools.ietf.org/html/rfc8555#section-7.3.5) but the usual 90-day renewal won't involve it AFAIK.

packager/util/config.go
```diff
@@ -24,13 +24,17 @@ import (
...
+	AutoRenewCert	bool   // Should we auto-renew cert? Defaults to false.
```
twifkak: (opt) Could we simplify? Auto-renew if and only if there's a non-empty AccountURL?

packager/util/config.go
```diff
@@ -50,6 +54,16 @@ type URLPattern struct {
 	SamePath               *bool
 }
 
+type ACMEConfig struct {
+	Prod	*ACMEServerConfig
+	Staging	*ACMEServerConfig
```
twifkak: I'd rename to Development; matches our `-development` command-line flag. We have plans for a `-staging` flag (#188) but I think that'll use production config in every way except the cert-url it writes into the SXG.

packager/util/config.go
banaag: NewCertFile will be read/write.  I envision that both CertFile and NewCertFile will both be set when both are valid (overlap) and that once CertFile becomes invalid, NewCertFile will replace it (CertFile = NewCertFile) and NewCertFile will be empty.  I would think that this applies to the disk copies as well (which may need to be some sort of shared disk, if multiple replicas of amppackager are running).

packager/util/config.go
banaag: That sounds reasonable.  I will adjust the comments to reflect that decision.

packager/util/config.go
banaag: Development sounds reasonable.

packager/util/config.go
banaag: You are correct, I will remove the field.

packager/util/config.go
twifkak: OK, SGTM! Thanks for clarifying.

packager/util/config.go
```diff
@@ -28,9 +28,13 @@ type Config struct {
 	Port		int
 	CertFile	string // This must be the full certificate chain.
 	KeyFile		string // Just for the first cert, obviously.
+
+	// NewCertFile will be read/write. CertFile and NewCertFile will be set when both
```
twifkak: How about for the first sentence, "When set, both CertFile and NewCertFile will be read/write." (Though, as you're probably aware, the user-facing documentation is in amppkg.example.toml.)

banaag: Done.

PR #339, opened 2019-08-22
Added utility func to compute cert expiry.

banaag:
Addresses #93

### DIFF
--- a/packager/util/util.go
+++ b/packager/util/util.go
@@ -78,14 +78,17 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 
 // Returns the Duration of time before cert expires with given deadline.
 // Note that the certExpiryDeadline should be the expected SXG expiration time.
-// Returns -1 if cert is already expired. This will be used to periodically check if cert
+// Returns 0 if cert is already expired. This will be used to periodically check if cert
 // is still within validity range.
 func GetDurationToExpiry(cert *x509.Certificate, certExpiryDeadline time.Time) (time.Duration, error) {
-	if cert.NotBefore.After(certExpiryDeadline) || cert.NotAfter.Before(certExpiryDeadline) {
-		return time.Duration(-1), errors.New("Certificate has invalid duration")
+	if cert.NotBefore.After(certExpiryDeadline) {
+		return 0, errors.New("Certificate is future-dated")
+	}
+	if cert.NotAfter.Before(certExpiryDeadline) {
+		return 0, errors.New("Certificate is expired")
 	}
 
-	return cert.NotAfter.Sub(certExpiryDeadline), nil;
+	return cert.NotAfter.Sub(certExpiryDeadline), nil
 }
 
 // CanSignHttpExchanges returns nil if the given certificate has the
--- a/packager/util/util.go
+++ b/packager/util/util.go
@@ -78,7 +78,7 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 
 // Returns the Duration of time before cert expires with given deadline.
 // Note that the certExpiryDeadline should be the expected SXG expiration time.
-// Returns 0 if cert is already expired. This will be used to periodically check if cert
+// Returns error if cert is already expired. This will be used to periodically check if cert
 // is still within validity range.
 func GetDurationToExpiry(cert *x509.Certificate, certExpiryDeadline time.Time) (time.Duration, error) {
 	if cert.NotBefore.After(certExpiryDeadline) {
--- a/packager/util/util.go
+++ b/packager/util/util.go
@@ -77,6 +77,7 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 }
 
 // Returns the Duration of time before cert expires with given deadline.
+// Note that the certExpiryDeadline should be the expected SXG expiration time.
 // Returns -1 if cert is already expired. This will be used to periodically check if cert
 // is still within validity range.
 func GetDurationToExpiry(cert *x509.Certificate, certExpiryDeadline time.Time) (time.Duration, error) {
--- a/packager/util/util.go
+++ b/packager/util/util.go
@@ -79,12 +79,12 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 // Returns the Duration of time before cert expires with given deadline.
 // Returns -1 if cert is already expired. This will be used to periodically check if cert
 // is still within validity range.
-func GetDurationToExpiry(cert *x509.Certificate, currentTime time.Time) (time.Duration, error) {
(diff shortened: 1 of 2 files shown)


packager/util/util.go
```diff
@@ -75,6 +76,17 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 	return false
 }
 
+// Returns the Duration of time before cert expires with given deadline.
+// Returns -1 if cert is already expired. This will be used to periodically check if cert
+// is still within validity range.
+func GetDurationToExpiry(cert *x509.Certificate, currentTime time.Time) time.Duration {
+	if cert.NotBefore.After(currentTime) || cert.NotAfter.Before(currentTime) {
```
twifkak: Depending on how this ends up being used, I'd suggest two changes:
1. Drop the `NotAfter.Before(currentTime)` bit; the return statement below will already return a negative value in that case.
2. Print a warning to the screen when `NotBefore.After(currentTime)`; this probably indicates some configuration error (either with the CA, or the local clock), and it would be good to have some knowledge if the server ends up spinning in a loop because of this.

If you choose to do #2 in the caller function instead of here (reasonable choice either way), then I'd prefer returning (time.Duration, error) instead of using a sentinel value.

packager/util/util_test.go
```diff
@@ -22,6 +23,26 @@ func TestCertName(t *testing.T) {
...
+	// Time after the cert validity.
+	afterCert := time.Date(2019, time.August, 8, 0, 0, 0, 0, time.UTC)
+	// Time 2 days before cert validity expiration.
+	twoDaysBeforeExpiry := time.Date(2019, time.August, 5, 5, 43, 32, 0, time.UTC)
+	// Time 0 days, 1 hour before cert validity expiration.
+	zeroDaysBeforeExpiry := time.Date(2019, time.August, 7, 4, 43, 32, 0, time.UTC)
+
+	assert.Equal(t, time.Duration(-1), util.GetDurationToExpiry(pkgt.B3Certs[0], beforeCert))
+	assert.Equal(t, time.Duration(-1), util.GetDurationToExpiry(pkgt.B3Certs[0], afterCert))
+	assert.Equal(t, time.Duration(2), util.GetDurationToExpiry(
```
twifkak: I'd suggest changing it to be `2 * 24 * time.Hour` so there's no loss of fidelity in the test (rounding due to integer division). Also, `time.Duration(2)` reads like "2 nanoseconds" to me.

packager/util/util.go
banaag: I decided to keep #1 because the logic seems cleaner if I can just return a valid duration plus nil (for the error).
I also decided to let the caller handle the error message.

shigeki: I think that it is no use to check if the cert is expired at the `currentTime` because the cert should be valid until the SXG signature is expired. It is too late to know its expiry at the `currentTime`.
The signature expiry time of `Expires:     now.Add(6 * 24 * time.Hour)` needs to be considered and it is better to add it in this check.

packager/util/util.go
```diff
@@ -75,6 +76,17 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 	return false
 }
 
+// Returns the Duration of time before cert expires with given deadline.
+// Returns -1 if cert is already expired. This will be used to periodically check if cert
+// is still within validity range.
+func GetDurationToExpiry(cert *x509.Certificate, currentTime time.Time) (time.Duration, error) {
+	if cert.NotBefore.After(currentTime) || cert.NotAfter.Before(currentTime) {
+		return time.Duration(-1), errors.New("Certificate has invalid duration")
```
twifkak: Go convention is to return the zero value on error. In this case, I think you can say `return 0, errors...`; the cast to `Duration` is implicit.

packager/util/util.go
```diff
@@ -75,6 +76,17 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
...
+	}
+
+	return cert.NotAfter.Sub(currentTime), nil;
```
twifkak: No semicolon

packager/util/util.go
twifkak: "invalid duration" error message is wrong. Perhaps "Certificate is expired". (Maybe have a second error message for `NotBefore` like "Certificate is future-dated".)

packager/util/util.go
banaag: I agree, this utility is not meant to check it at current time (as in time.Now()), it was meant to check the cert against a specified deadline.  Your point about the naming of the parameters is well taken, I will rename it to certExpiryDeadline.

packager/util/util_test.go
```diff
@@ -22,6 +23,33 @@ func TestCertName(t *testing.T) {
...
+	afterCert := time.Date(2019, time.August, 8, 0, 0, 0, 0, time.UTC)
+	// Time 2 days before cert validity expiration.
+	twoDaysBeforeExpiry := time.Date(2019, time.August, 5, 5, 43, 32, 0, time.UTC)
+	// Time 0 days, 1 hour before cert validity expiration.
+	oneHourBeforeExpiry := time.Date(2019, time.August, 7, 4, 43, 32, 0, time.UTC)
+	// Time 0 days before cert validity expiration.
+	zeroDaysBeforeExpiry := time.Date(2019, time.August, 7, 5, 43, 32, 0, time.UTC)
+
+	d, err := util.GetDurationToExpiry(pkgt.B3Certs[0], beforeCert)
+	assert.Contains(t, errorFrom(err), "Certificate has invalid duration")
```
twifkak: Maybe use https://godoc.org/github.com/stretchr/testify/assert#EqualError ? Unless an exact string match is prohibitive.

packager/util/util.go
twifkak: Maybe add a note, too, to the function doc that this should be the expected SXG expiration time.

@shigeki Also note that we discussed making a toml config parameter (something like renewalGracePeriod), with a default of say, 2 days, so that the process for requesting a new cert via ACME will begin 6+2=8 days before expiration, to allow for ACME server outages. Does this seem OK?

shigeki: c365be83b9490e95ba4b46de67c58b0b294da0d3 is fine for me. Thanks.

shigeki:
@twifkak 8 days are fine because OCSP next update is 7 days duration and we need not care about OCSP validation period against the cert expiry date.

packager/util/util.go
```diff
@@ -78,14 +78,17 @@ func hasCanSignHttpExchangesExtension(cert *x509.Certificate) bool {
 
 // Returns the Duration of time before cert expires with given deadline.
 // Note that the certExpiryDeadline should be the expected SXG expiration time.
-// Returns -1 if cert is already expired. This will be used to periodically check if cert
+// Returns 0 if cert is already expired. This will be used to periodically check if cert
```
twifkak: s/0/error/

PR #347, opened 2019-09-17
Sync from Google

twifkak:
This is an automatically generated pull request.

PR #348, opened 2019-09-17
Limit signature duration based on max_age_secs.

twifkak:
This uses the new metadata field returned from the transformer library,
which is based on the presence of any inline amp-scripts.

@Gregable Note for review to ignore the first commit; it's merely the "diffbase" of #347.

cc @choumx

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -122,15 +122,15 @@ type Signer struct {
 	// TODO(twifkak): Support multiple certs. This will require generating
 	// a signature for each one. Note that Chrome only supports 1 signature
 	// at the moment.
-	certHandler	certcache.CertHandler
+	certHandler certcache.CertHandler
 	// TODO(twifkak): Do we want to allow multiple keys?
-	key             crypto.PrivateKey
-	client          *http.Client
-	urlSets         []util.URLSet
-	rtvCache        *rtv.RTVCache
-	shouldPackage   func() bool
-	overrideBaseURL *url.URL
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -477,11 +477,17 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 	if err != nil {
 		util.NewHTTPError(http.StatusInternalServerError, "Error building validity href: ", err).LogAndRespond(resp)
 	}
+	// Expires - Date must be <= 604800 seconds, per
+	// https://tools.ietf.org/html/draft-yasskin-httpbis-origin-signed-exchanges-impl-00#section-3.5.
+	duration := 7*24*time.Hour
```
Gregable: suggest spaces around * for readability.

packager/signer/signer.go
```diff
@@ -477,11 +477,17 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
...
+	println("max-age=", metadata.MaxAgeSecs)
```
Gregable: debug statement?

packager/signer/signer.go
twifkak: Okay, ran go fmt on these two files, which fixed this.

packager/signer/signer.go
twifkak: Oops, removed.

PR #349, opened 2019-09-18
Integrate CertFetcher with flag-protection into AMP Packager.

banaag:
Update certcache to use certfetcher if cert autorenew is turned on.
Update ACME config to include email adddress and acme challenge port.
Update certloader.PopulateCertCache to instantiate certfetcher and pass it on to certcache instance.
Update all relevant tests.


packager/certcache/certcache.go
```diff
@@ -46,10 +47,19 @@ var infiniteFuture = time.Unix(1<<63-62135596801, 999999999)
...
+const MaxOCSPResponseBytes = 1024 * 1024
 
 // How often to check if OCSP stapling needs updating.
-const ocspCheckInterval = 1 * time.Hour
+const OcspCheckInterval = 1 * time.Hour
+
+// How often to check if certs needs updating.
+const CertCheckInterval = 24 * time.Hour
+
+// Recommended renewal duration for certs.
```
Gregable: Maybe clarify in the comments that this is the duration before the next expiry rather than  the duration after the last renewal.

packager/certcache/certcache.go
```diff
@@ -122,15 +143,40 @@ func (this *CertCache) Init(stop chan struct{}) error {
...
-	// TODO(banaag): check if cert is still valid, refresh if not.
-	return this.certs[0]
+	if !this.isInitialized || this.certFetcher == nil {
+		// If certcache is not initialized or certFetcher is not set,
+		// just return cert without checking if it needs auto-renewal.
+		return this.certs[0]
+	}
+	d, err := util.GetDurationToExpiry(this.certs[0], this.certs[0].NotAfter)
+	if err != nil {
+		// Current cert is already invalid. Check if renewal is available.
```
Gregable: Consider using TODO to mark places for the follow-on changes.

packager/certcache/certcache.go
banaag: Done.

packager/certcache/certcache.go
banaag: I already try to do this, what you pointed out is code I missed updating and now fixed. Thanks!

packager/certcache/certcache.go
```diff
@@ -445,3 +570,171 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
...
+	d := time.Duration(0)
+	err := errors.New("")
+	if this.hasCert() {
+		d, err = util.GetDurationToExpiry(this.certs[0], time.Now())
+	}
+	if err != nil {
+		if this.renewedCerts != nil {
+			// If renewedCerts is set, copy that over to certs
+			// and set renewedCerts to nil.
+			// TODO(banaag): do the same cert setting dance on disk.
```
twifkak: Is this TODO still accurate? setCerts/setNewCerts seem to write to disk.

packager/certcache/certcache.go
```diff
@@ -445,3 +570,171 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
...
+			// Purge OCSP cache? Make shouldUpdateOCSP() return true?
```
twifkak: Ooh, don't forget TODO this (future PR OK).

packager/certcache/certcache.go
```diff
@@ -445,3 +570,171 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
...
+			log.Println("Error trying to fetch new certificates from CA: ", err)
+			return
+		}
+		this.setNewCerts(certs)
+	}
+}
+
+func (this *CertCache) reloadCertIfExpired() {
+	// We always validate the certs here.  If we are in development mode and the certs don't validate,
+	// it doesn't matter because the old certs won't be overriden (and the old certs are probably invalid, too).
```
twifkak: overridden

packager/certcache/certcache.go
```diff
@@ -58,8 +73,13 @@ type CertHandler interface {
...
 type CertCache struct {
 	// TODO(twifkak): Support multiple cert chains (for different domains, for different roots).
-	certName          string
-	certs             []*x509.Certificate
+	certName string
+	certsMu  sync.RWMutex
+	certs    []*x509.Certificate
+	// If certFetcher is not set, that means cert auto-renewal is not available.
+	certFetcher       *certfetcher.CertFetcher
+	renewedCertsMu    sync.RWMutex
```
twifkak: 1. The certName should change on a cert renewal (the sha256 computed by util.CertName is different), I think? So I think we need a CertCache.renewedCertName field also.
2. I think we need to put any reads/writes of certName and renewedCertName under their respective RWMutex.

packager/certcache/certcache.go
```diff
@@ -445,3 +570,171 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
...
+		// Cert is still valid, don't do anything.
+	} else if d < time.Duration(CertRenewalInterval) {
+		// Cert is still valid, but we need to start process of requesting new cert.
+		log.Println("Warning: Current cert crossed threshold for renewal, attempting to renew.")
+		certs, err := this.certFetcher.FetchNewCert()
+		if err != nil {
+			log.Println("Error trying to fetch new certificates from CA: ", err)
+			return
+		}
+		this.setNewCerts(certs)
```
twifkak: This logic seems incorrect to me; sorry if I read it wrong. AFAICT:

1. 8 days before expiry, this code runs and creates the new cert file. Nothing uses it yet.
2. 1 day later, maintainCerts calls this function again, which will lead down here to another call to FetchNewCert. Ditto daily.
3. Finally, maintainCerts calls this function and the old cert is expired, at which point the new cert is copied on top of it.

Step 2 seems like it'll make multiple redundant ACME requests for a cert.

Step 3 seems like it's happening at the wrong time (both too soon and too late). It seems like we should wait until we can get a `GOOD` OCSP response for it, and then immediately start serving it and using it to sign SXGs, so that we don't enter that 7-day period when we can't sign SXGs that last long enough.

packager/certloader/certloader.go
```diff
@@ -17,33 +17,89 @@ package certloader
...
+		return nil, errors.New("missing ACMEConfig")
+	}
+	if config.ACMEConfig.Production == nil {
+		return nil, errors.New("missing ACMEConfig.Production")
 	}
-	for _, urlSet := range config.URLSet {
-		domain := urlSet.Sign.Domain
-		if err := util.CertificateMatches(certs[0], key, domain); err != nil {
-			return nil, errors.Wrapf(err, "checking %s", config.CertFile)
+	if config.ACMEConfig.Production.EmailAddress == "" {
```
twifkak: Typically I've put this kind of validation in util/config.go. (Pro is the definition of what's a valid config is in one place; con is it's farther away from the thing actually using it.) WDYT?

go.sum
```diff
@@ -1,71 +1,102 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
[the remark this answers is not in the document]
banaag: The dependency mostly has to do with the fact that DNS validation requires various DNS provider/cloud library apis that get pulled in by the lego ACME library.

packager/certcache/certcache.go
```diff
@@ -46,10 +47,19 @@ var infiniteFuture = time.Unix(1<<63-62135596801, 999999999)
 // 1MB is the maximum used by github.com/xenolf/lego/acmev2 in GetOCSPForCert.
 // Alternatively, here's a random documented example of a 20K limit:
 // https://www.ibm.com/support/knowledgecenter/en/SSPREK_9.0.0/com.ibm.isam.doc/wrp_stza_ref/reference/ref_ocsp_max_size.html
-const maxOCSPResponseBytes = 1024 * 1024
+const MaxOCSPResponseBytes = 1024 * 1024
```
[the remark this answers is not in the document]
banaag: Oops, my bad.  I thought constants needed to be capitalized.  Re-read the style guide and it only applies to exports.  Fixed all the constants and also used maxOCSPResponseBytes in io.LimitReader() as per your intention.

packager/certcache/certcache.go
```diff
@@ -46,10 +47,19 @@ var infiniteFuture = time.Unix(1<<63-62135596801, 999999999)
...
 // How often to check if OCSP stapling needs updating.
-const ocspCheckInterval = 1 * time.Hour
+const OcspCheckInterval = 1 * time.Hour
+
+// How often to check if certs needs updating.
+const CertCheckInterval = 24 * time.Hour
+
+// Recommended renewal duration for certs. This is duration before next cert expiry.
+// 8 days is recommended duration to start requesting new certs to allow for ACME server outages.
+// It's 6 days + 2 days renewal grace period.
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -58,25 +68,39 @@ type CertHandler interface {
...
+	ocspFile	  Updateable
+	client		  http.Client
 
 	// "Virtual methods", exposed for testing.
 	// Given a certificate, returns the OCSP responder URL for that cert.
 	extractOCSPServer func(*x509.Certificate) (string, error)
 	// Given an HTTP request/response, returns its cache expiry.
 	httpExpiry func(*http.Request, *http.Response) time.Time
+	// Is CertCache initialized to do cert renewal or OCSP refreshes?
+	isInitialized	  bool
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -58,25 +68,39 @@ type CertHandler interface {
...
 }
 
-// Must call Init() on the returned CertCache before you can use it.
-func New(certs []*x509.Certificate, ocspCache string) *CertCache {
+// Callers need to call Init() on the returned CertCache before the cache can auto-renew certs.
+// Callers can use the uninitialized CertCache either for testing certificates (without doing OCSP or
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
banaag: I did the cert setting dance on disk.  The Purge OCSP cache can be a future PR.

packager/certcache/certcache.go
banaag: Done.

packager/certcache/certcache.go
banaag: Let me think about this and our offline discussion about OCSP checks and send a follow-up PR when it's done. Thanks!

packager/certcache/certcache.go
```diff
@@ -440,3 +503,96 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
...
+	defer this.renewedCertsMu.Unlock()
+	this.renewedCerts = certs
+}
+
+// Update the cert in the cache if necessary.
+func (this *CertCache) updateCertIfNecessary() {
+	log.Println("Updating cert if necessary");
+	if this.certFetcher == nil {
+		// Do nothing if certFetcher is not set.
+		log.Println("Certfetcher is not set, skipping cert updates.");
```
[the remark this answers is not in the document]
banaag: This is not possible in this code flow.  All I know here is that certfetcher is not set since it's nil.  The reason is not known at this point (it was known during an earlier flow).  The logging will just be useful for debugging as an indication that certificate fetching is not happening.

vendor/modules.txt
```diff
@@ -1,3 +1,31 @@
+# cloud.google.com/go v0.38.0
```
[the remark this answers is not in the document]
banaag: ls -l produces:
47048294 Nov  7 08:58  amppkg

packager/certcache/certcache.go
```diff
@@ -58,25 +68,39 @@ type CertHandler interface {
...
 }
 
-// Must call Init() on the returned CertCache before you can use it.
-func New(certs []*x509.Certificate, ocspCache string) *CertCache {
+// Callers need to call Init() on the returned CertCache before the cache can auto-renew certs.
+// Callers can use the uninitialized CertCache either for testing certificates (without doing OCSP or
+// cert refreshes).
+func New(certs []*x509.Certificate, certFetcher *certfetcher.CertFetcher, ocspCache string) *CertCache {
+	certName := ""
+	if len(certs) > 0 && certs[0] != nil {
```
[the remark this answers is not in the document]
banaag: Can't use it here, the instance variable is not set yet.  We are checking the argument of New(), not this.certs.

packager/certfetcher/certfetcher.go
```diff
@@ -93,28 +126,23 @@ func NewFetcher(email string, privateKey crypto.PrivateKey, acmeDiscoURL string,
...
 func (f *CertFetcher) FetchNewCert() ([]*x509.Certificate, error) {
-	request := certificate.ObtainRequest{
-		Domains: f.Domains,
-		Bundle:  true,
-	}
-
 	// Each resource comes back with the cert bytes, the bytes of the client's
 	// private key, and a certificate URL.
-	resource, err := f.legoClient.Certificate.Obtain(request)
+	resource, err := f.legoClient.Certificate.ObtainForCSR(*(f.CertSignRequest), true)
```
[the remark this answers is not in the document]
banaag: Done.

packager/certfetcher/certfetcher.go
```diff
@@ -54,8 +55,10 @@ func (u *AcmeUser) GetPrivateKey() crypto.PrivateKey {
 }
 
 // Initializes the cert fetcher with information it needs to fetch new certificates in the future.
-func NewFetcher(email string, privateKey crypto.PrivateKey, acmeDiscoURL string,
-	domains []string, acmeChallengePort int, shouldRegister bool) (*CertFetcher, error) {
+func NewFetcher(email string, certSignRequest *x509.CertificateRequest, privateKey crypto.PrivateKey,
```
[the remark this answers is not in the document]
banaag: Done.

packager/certfetcher/certfetcher.go
```diff
@@ -71,14 +74,44 @@ func NewFetcher(email string, privateKey crypto.PrivateKey, acmeDiscoURL string,
...
+	// We specify an http port of `httpChallengePort`
 	// because we aren't running as root and can't bind a listener to port 80 and 443
 	// (used later when we attempt to pass challenges). Keep in mind that you still
 	// need to proxy challenge traffic to port `acmeChallengePort`.
-	err = client.Challenge.SetHTTP01Provider(
-		http01.NewProviderServer("", strconv.Itoa(acmeChallengePort)))
-	if err != nil {
-		return nil, errors.Wrap(err, "Setting up HTTP01 challenge provider.")
+	if httpChallengePort != 0 {
+		err = client.Challenge.SetHTTP01Provider(
```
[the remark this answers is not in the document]
banaag: Done.

packager/certfetcher/certfetcher.go
```diff
@@ -71,14 +74,44 @@ func NewFetcher(email string, privateKey crypto.PrivateKey, acmeDiscoURL string,
...
+			http01.NewProviderServer("", strconv.Itoa(httpChallengePort)))
+		if err != nil {
+			return nil, errors.Wrap(err, "Setting up HTTP01 challenge provider.")
+		}
+	}
+	if httpChallengeWebRoot != "" {
```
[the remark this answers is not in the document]
banaag: Yes, it supports multiple challenge types and tries them depending on what's been made available.  If the needed challenge types are not available, the validation fails.

packager/certloader/certloader.go
```diff
@@ -17,33 +17,89 @@ package certloader
...
+			config.ACMEConfig.Development.DnsProvider == "" {
+			return nil, errors.New("One of HttpChallengePort, HttpWebRootDir, TlsChallengePort and DnsProvider must be present.")
+		}
+		httpChallengePort = config.ACMEConfig.Development.HttpChallengePort
+		httpWebRootDir = config.ACMEConfig.Development.HttpWebRootDir
+		tlsChallengePort = config.ACMEConfig.Development.TlsChallengePort
+		dnsProvider = config.ACMEConfig.Development.DnsProvider
+	}
+
+	csr, err := LoadCSRFromFile(config)
```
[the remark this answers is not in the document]
banaag: Done.

packager/certloader/certloader.go
```diff
@@ -55,21 +111,45 @@ func PopulateCertCache(config *util.Config, key crypto.PrivateKey, developmentMo
 //	 (if developmentMode, print a warning that certs can't
 //	 be used to sign HTTP exchanges).
 // If there are no errors, the array of certificates is returned.
-func loadCertsFromFile(config *util.Config, developmentMode bool) ([]*x509.Certificate, error) {
+func LoadCertsFromFile(config *util.Config, developmentMode bool) ([]*x509.Certificate, error) {
+	return LoadAndValidateCertsFromFile(config.CertFile, !developmentMode)
+}
+
+func LoadAndValidateCertsFromFile(certPath string, checkIfCanSign bool) ([]*x509.Certificate, error) {
```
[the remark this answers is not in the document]
banaag: Done.

packager/certloader/certloader.go
```diff
@@ -55,21 +111,45 @@ func PopulateCertCache(config *util.Config, key crypto.PrivateKey, developmentMo
...
+	// Use independent .lock file; necessary on Windows to avoid "The process cannot
+	// access the file because another process has locked a portion of the file."
+	lockPath := certPath + ".lock"
+	lock := flock.New(lockPath)
+	locked, err := lock.TryLock()
```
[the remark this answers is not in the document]
banaag: Done.

packager/certloader/certloader.go
```diff
@@ -79,6 +159,91 @@ func loadCertsFromFile(config *util.Config, developmentMode bool) ([]*x509.Certi
...
+			log.Printf("Error unlocking %s; %+v", lockPath, err)
+		}
+		if err := os.Remove(lockPath); err != nil {
+			log.Printf("Error removing %s; %+v", lockPath, err)
+		}
+	}()
+
+	cert := certToPEM(certs[0])
+	issuer := certToPEM(certs[1])
+	bundled := append(cert, issuer...)
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -46,10 +49,22 @@ var infiniteFuture = time.Unix(1<<63-62135596801, 999999999)
...
+const MaxOCSPResponseBytes = 1024 * 1024
 
 // How often to check if OCSP stapling needs updating.
-const ocspCheckInterval = 1 * time.Hour
+const OcspCheckInterval = 1 * time.Hour
+
+// How often to check if certs needs updating.
+const CertCheckInterval = 24 * time.Hour
+
+// Max number of OCSP request tries.
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -445,3 +570,171 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
 	}
 	return respBytes
 }
+
+// Checks for cert updates every certCheckInterval hours. Never terminates.
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -381,7 +505,7 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
 	// StdEncoding).
 	getURL := ocspServer + "/" + url.PathEscape(base64.StdEncoding.EncodeToString(req))
 	var httpReq *http.Request
-	if len(getURL) <= 255 {
+	if len(getURL) <= 255 && !isRetry {
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
banaag: Removed.

packager/certloader/certloader.go
```diff
@@ -17,33 +17,89 @@ package certloader
...
+	developmentMode bool, autoRenewCert bool) (*certfetcher.CertFetcher, error) {
+	if !autoRenewCert {
+		// Certfetcher can be nil, if auto renew is off.
+		return nil, nil
+	}
+
+	if config.ACMEConfig == nil {
+		return nil, errors.New("missing ACMEConfig")
+	}
+	if config.ACMEConfig.Production == nil {
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -440,3 +503,96 @@ func (this *CertCache) fetchOCSP(orig []byte, ocspUpdateAfter *time.Time) []byte
...
+func (this *CertCache) getCert() *x509.Certificate {
+	if !this.hasCert() {
+		return nil
+	}
+	return this.certs[0]
+}
+
+// Set current cert with mutex protection.
+func (this *CertCache) setCerts(certs []*x509.Certificate) {
+	this.certsMu.Lock()
```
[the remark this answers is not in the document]
banaag: certsMU is already an RWMutex (you maybe looking at an older version of the code when you commented), added RLock-protect for all reads also. Did the same for renewedCerts.

packager/certcache/certcache.go
```diff
@@ -235,18 +310,52 @@ func (this *CertCache) isHealthy(ocspResp []byte) error {
...
+	for numTries := 0; numTries < maxTries; {
+		ocsp, err = this.ocspFile.Read(context.Background(), this.shouldUpdateOCSP, func(orig []byte) []byte {
+			return this.fetchOCSP(orig, &ocspUpdateAfter, numTries > 0)
+		})
+		if err != nil {
+			if numTries >= maxTries-1 {
+				return nil, time.Time{}, errors.Wrap(err, "Updating OCSP cache")
+			} else {
+				numTries++
+				waitTimeInMinutes = waitForSpecifiedTime(waitTimeInMinutes, numTries)
```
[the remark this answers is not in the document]
banaag: Added the allowRetries arg to readOCSP.  Added TODO for retrywithBackoff idea.

banaag:
Fixed the comments I missed (thanks twifkak@). Please feel free to continue review. @twifkak  @Gregable

packager/certloader/certloader.go
```diff
@@ -184,9 +174,14 @@ func WriteCertsToFile(certs []*x509.Certificate, filepath string) error {
 		}
 	}()
 
-	cert := certToPEM(certs[0])
-	issuer := certToPEM(certs[1])
-	bundled := append(cert, issuer...)
+	certLen := len(certs)
+	bundled := make([]byte, 0, 0)
```
twifkak: Just `[]byte{}` perhaps?

packager/certloader/certloader.go
```diff
@@ -184,9 +174,14 @@ func WriteCertsToFile(certs []*x509.Certificate, filepath string) error {
...
+	for i := 0; i < certLen - 1; i++ {
```
twifkak: Can you just change `certLen - 1` to `certLen` and get rid of the special case `issuer` after the loop? Better still use `for _, cert := range certs {`.

packager/certloader/certloader.go
```diff
@@ -184,9 +174,14 @@ func WriteCertsToFile(certs []*x509.Certificate, filepath string) error {
...
+		cert := certToPEM(certs[i])
+		bundled = append(bundled, cert...)
+	}
+	issuer := certToPEM(certs[certLen])
```
twifkak: Won't `certs[certLen]` cause an out-of-range error?

Also, FYI, other than the leaf cert being at position 0, there's no ordering requirement for the rest of the certs array. The issuer cert can be any of the subsequent certs. (Hence why findIssuer() exists.)

packager/certcache/certcache.go
```diff
@@ -669,30 +750,64 @@ func (this *CertCache) updateCertIfNecessary() {
...
+			certs, err := this.certFetcher.FetchNewCert()
+			if err != nil {
+				log.Println("Error trying to fetch new certificates from CA: ", err)
+				return
+			}
+			this.setNewCerts(certs)
+		} else {
+			var ocspUpdateAfter time.Time
+
+			ocsp, _, errorOcsp := this.readOCSP(true)
```
twifkak: It's Go style to capitalize the initialisms inside camel-case, e.g. errorOCSP and newOCSP. (But keep it lowercase if it's the first one, like ocspUpdateAfter.)

packager/certcache/certcache.go
```diff
@@ -669,30 +750,64 @@ func (this *CertCache) updateCertIfNecessary() {
...
+			if errorOcsp != nil {
+				newOcsp := this.fetchOCSP(ocsp, this.renewedCerts, &ocspUpdateAfter, false)
```
twifkak: AFAICT, this is going to try to fetch the OCSP response for the new cert once a day, rather than using the backoff retry logic you added to readOCSP. Is that desired? Maybe that's OK for now, but can you add a TODO to revisit?

I realize it's difficult to use readOCSP here, since it's hard-coded to use this.certs and friends, rather than this.renewedCerts and friends. That makes me think two things:
1. The extraction of the retry logic from readOCSP would be useful.
2. We should bundle certName, certs, certsMu, ocspFile, ocspFilePath, ocspUpdateAfter, and ocspUpdateAfterMu into a new struct type, and then have two copies of that in certcache - one for current certs and one for new certs.

OK for future PR... TODO? :sheep: :grin:

packager/certcache/certcache.go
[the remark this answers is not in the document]
banaag: Oops, fixed.

packager/certcache/certcache.go
```diff
@@ -234,6 +237,9 @@ func (this *CertCache) ocspMidpoint(bytes []byte, issuer *x509.Certificate) (tim
 
 func (this *CertCache) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 	params := mux.Params(req)
+
+	this.certsMu.Lock()
```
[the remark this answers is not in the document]
banaag: This is already RLock and RUnlock in the current version (maybe you were looking at an older file?).  I did move it closer to the read now, as you recommended.

packager/certcache/certcache.go
```diff
@@ -475,22 +481,50 @@ func (this *CertCache) findIssuer() *x509.Certificate {
 	return nil
 }
 
+// Finds the issuer of the specified cert (i.e. the second from the bottom of the
+// chain).
+func (this *CertCache) findIssuerUsingCerts(certs []*x509.Certificate) *x509.Certificate {
```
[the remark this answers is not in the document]
banaag: Done.

packager/certcache/certcache.go
```diff
@@ -669,30 +718,61 @@ func (this *CertCache) updateCertIfNecessary() {
 		this.setCerts(certs)
```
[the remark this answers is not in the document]
banaag: Good catch. Done.

packager/certcache/certcache.go
```diff
@@ -669,30 +718,61 @@ func (this *CertCache) updateCertIfNecessary() {
...
-		this.setNewCerts(certs)
 	}
 }
 
 func (this *CertCache) reloadCertIfExpired() {
+        d := time.Duration(0)
+        err := errors.New("")
+        if this.hasCert() {
+                d, err = util.GetDurationToExpiry(this.certs[0], time.Now())
+        }
```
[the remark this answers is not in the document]
banaag: Ran go fmt certcache.go

packager/certcache/certcache.go
```diff
@@ -669,30 +718,61 @@ func (this *CertCache) updateCertIfNecessary() {
...
+				}
+			}
 		}
-		this.setNewCerts(certs)
 	}
 }
 
 func (this *CertCache) reloadCertIfExpired() {
+        d := time.Duration(0)
+        err := errors.New("")
```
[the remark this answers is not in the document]
banaag: Done.

packager/certloader/certloader.go
```diff
@@ -40,60 +40,50 @@ func CreateCertFetcher(config *util.Config, key crypto.PrivateKey, domain string
...
-		httpWebRootDir = config.ACMEConfig.Development.HttpWebRootDir
-		tlsChallengePort = config.ACMEConfig.Development.TlsChallengePort
-		dnsProvider = config.ACMEConfig.Development.DnsProvider
-	}
+	httpChallengePort := acmeConfig.HttpChallengePort
+	httpWebRootDir := acmeConfig.HttpWebRootDir
+	tlsChallengePort := acmeConfig.TlsChallengePort
+	dnsProvider := acmeConfig.DnsProvider
 
+	// TODO(banaag): Rather than making them create a CSR, generate one using the given KeyFile/CertFile and
```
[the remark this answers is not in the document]
banaag: Done.

packager/certloader/certloader.go
banaag: Done.

packager/certloader/certloader.go
banaag: Done.

packager/certloader/certloader.go
banaag: Ok fixed in previous comment.  Good to know about issuer position.

packager/certcache/certcache.go
banaag: Done.

packager/certcache/certcache.go
banaag: The logic for the certs is by design and is as follows:
updateCertIfNecessary() will be called:
1. On demand, during initialization or when GetLatestCert() is called and we already have an expired cert.
2. Every 'certCheckInterval' hours (currently every 24 hours).

The logic for retries in readOCSP is specific to just reading OCSP and retrying when we got a new cert because of Digicert's setup.

Inside updateCertIfNecessary, there's a case where we already got the renewalCert (waiting in this.renewedCerts) and in this case, we make sure we also get the OCSP by calling readOCSP with retries set to true.

I understand your suggestion of possibly reusing the readOCSP retry logic in this edge case, but given that the updateCertIfNecessary() calls are a day apart, chances are the fetchOCSP might fail the first time, but will succeed the next day.  Given that, it will still be nice-to-have the refactor of readOCSP so I added this as a TODO.

packager/certcache/certcache.go
```diff
@@ -233,20 +319,63 @@ func (this *CertCache) isHealthy(ocspResp []byte) error {
...
+		if err != nil {
+			if numTries >= maxTries-1 {
+				return nil, time.Time{}, errors.Wrap(err, "Updating OCSP cache")
+			} else {
+				numTries++
+				waitTimeInMinutes = waitForSpecifiedTime(waitTimeInMinutes, numTries)
+				continue
+			}
+		}
+		if len(ocsp) == 0 {
```
Gregable: You have 3 places in this method that catch different error cases and trigger the numTries++;continue logic.

Consider extracting a method that does the ocsp read, checks all of the conditions (read error, empty file, unhealthy ocsp) and returns an error on any. This method and for loop then just implements the retry on error logic, but only needs a single condition/continue block.

packager/certcache/certcache.go
```diff
@@ -233,20 +319,63 @@ func (this *CertCache) isHealthy(ocspResp []byte) error {
...
+				return nil, time.Time{}, errors.Wrap(err, "OCSP failed health check.")
+			} else {
+				numTries++
+				waitTimeInMinutes = waitForSpecifiedTime(waitTimeInMinutes, numTries)
+				continue
+			}
+		} else {
+			break
+		}
+		numTries++
```
Gregable: Is this reachable?

The previous condition block starting l.367 always exists I think.

packager/certcache/certcache.go
```diff
@@ -256,9 +385,23 @@ func (this *CertCache) readOCSP() ([]byte, time.Time, error) {
...
+
+}
+
+// Print # of retries, wait for specified time and returned updated wait time.
+func waitForSpecifiedTime(waitTimeInMinutes int, numRetries int) int {
+	log.Printf("Retrying OCSP server: retry #%d\n", numRetries)
+	// Wait using exponential backoff.
+	log.Printf("Waiting for %d minute(s)\n", waitTimeInMinutes)
+	waitTimeDuration := time.Duration(waitTimeInMinutes) * time.Minute
+	// For exponential backoff.
```
Gregable: There is no "right" read pattern, so feel free to ignore as bikeshedding.

I think that the cost of making an OCSP request to the CA is reasonably low enough that we could do it every minute or 10 until it succeeds or we time out.

With this backoff, the last lookup is 8hrs apart? Seems excessively long to wait. I'd consider something like every 30s for 5 minutes, and then every 5 minutes for the next 12 hrs.

packager/certfetcher/certfetcher.go
```diff
@@ -54,8 +55,10 @@ func (u *AcmeUser) GetPrivateKey() crypto.PrivateKey {
 }
 
 // Initializes the cert fetcher with information it needs to fetch new certificates in the future.
-func NewFetcher(email string, privateKey crypto.PrivateKey, acmeDiscoURL string,
-	domains []string, acmeChallengePort int, shouldRegister bool) (*CertFetcher, error) {
+func New(email string, certSignRequest *x509.CertificateRequest, privateKey crypto.PrivateKey,
+	acmeDiscoURL string, httpChallengePort int, httpChallengeWebRoot string,
+	tlsChallengePort int, dnsProvider string, shouldRegister bool) (*CertFetcher, error) {
```
Gregable: optional again. Same thought here about breaking this up. Callsite could have some structure like:

fetcher := CertFetcher()
fetcher.setUser(email, privateKey)
fetcher.bindToPort(port)

packager/certcache/certcache.go
banaag: Added TODO.

packager/certcache/certcache.go
```diff
@@ -73,11 +104,19 @@ type CertCache struct {
 	httpExpiry func(*http.Request, *http.Response) time.Time
 }
 
-// Must call Init() on the returned CertCache before you can use it.
-func New(certs []*x509.Certificate, ocspCache string) *CertCache {
+// Callers need to call Init() on the returned CertCache before the cache can auto-renew certs.
+// Callers can use the uninitialized CertCache for testing certificates (without doing OCSP or
+// cert refreshes).
+func New(certs []*x509.Certificate, certFetcher *certfetcher.CertFetcher, domains []string,
+	certFile string, newCertFile string, ocspCache string) *CertCache {
```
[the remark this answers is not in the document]
banaag: Added TODO.

packager/certfetcher/certfetcher.go
banaag: Added TODO.

packager/certcache/certcache.go
banaag: Done.

packager/certcache/certcache.go
banaag: Fixed with refactor from comment above.

packager/certcache/certcache.go
banaag: I changed the logic so that it caps the wait time at 10 minutes, max.

banaag:
Fixed @gregable comments. PTAL.

packager/certcache/certcache.go
```diff
@@ -107,6 +107,12 @@ type CertCache struct {
 // Callers need to call Init() on the returned CertCache before the cache can auto-renew certs.
 // Callers can use the uninitialized CertCache for testing certificates (without doing OCSP or
 // cert refreshes).
+//
+// TODO(banaag): per greigable@ comments:
```
twifkak: *gregable

packager/certcache/certcache.go
```diff
@@ -318,6 +324,39 @@ func (this *CertCache) isHealthy(ocspResp []byte) error {
...
+			return nil, time.Time{}, nil
+		}
+	}
+
+	return ocsp, ocspUpdateAfter, nil
+}
+
 // Returns the OCSP response and expiry, refreshing if necessary.
 // TODO(banaag): Per twifkak's suggestion, consider:
 // It may also be interesting to try to separate the retry logic from the fetch logic. One approach comes to mind:
```
twifkak: I think you just did this TODO, essentially?

packager/certcache/certcache.go
```diff
@@ -396,6 +411,10 @@ func waitForSpecifiedTime(waitTimeInMinutes int, numRetries int) int {
 	waitTimeDuration := time.Duration(waitTimeInMinutes) * time.Minute
 	// For exponential backoff.
 	newWaitTimeInMinutes := 2 * waitTimeInMinutes
+	if newWaitTimeInMinutes > 10 {
+		// Cap the wait time at 10 minutes.
+		newWaitTimeInMinutes = 10
```
twifkak: Per this change, the comment on maxOCSPTries needs to be updated.

packager/certcache/certcache.go
banaag: Will address in follow-up PR #369

packager/certcache/certcache.go
banaag: Will address in follow-up PR #369

packager/certcache/certcache.go
banaag: Will address in follow-up PR #369

PR #360, opened 2019-10-23
Fix certcache to allow self-signed certs.

### LINKED ISSUE #352: Development mode requires a cert with an OCSP URL
Using the [instructions](https://github.com/WICG/webpackage/tree/master/go/signedexchange) to generate a self-signed cert, this doesn't work with `amppkg -development`:

```
2019/10/09 07:33:59 Updating OCSP; none cached yet.
2019/10/09 07:33:59 Updating OCSP; none cached yet.
2019/10/09 07:33:59 Error computing OCSP midpoint: Parsing OCSP: no response matching the supplied certificate
2019/10/09 07:33:59 Error computing OCSP midpoint: Parsing OCSP: no response matching the supplied certificate
2019/10/09 07:33:59 Error extracting OCSP server: Cert missing OCSPServer.
2019/10/09 07:33:59 Error parsing OCSP response: no response matching the supplied certificate
2019/10/09 07:33:59 OCSP failed health check.
```

We should fix `amppkg -development` not to require an OCSP URL, and in the meantime, document how to generate a self-signed cert with a fake AIA extension.

twifkak:
If the cert doesn't have an AIA extension with an OCSP responder URL,
then the cache will just include a fake (invalid) OCSP response in its
cert-chain+cbor. This is sufficient for displaying the SXG in Chrome
with --ignore-certificate-errors-spki-list.

Fixes #352.

### DIFF
--- a/packager/certcache/certcache.go
+++ b/packager/certcache/certcache.go
@@ -69,6 +69,10 @@ const maxOCSPTries = 10
 // TODO(banaag): make 2 days renewal grace period configurable in toml.
 const certRenewalInterval = 8 * 24 * time.Hour
 
+// Sentinel value used to communicate that a returned OCSP was fake, and the caller should not attempt to parse it.
+// No, I'm not proud.
+var fakeOCSP = []byte("fake ocsp response")
+
 type CertHandler interface {
 	GetLatestCert() *x509.Certificate
 	IsHealthy() error
@@ -90,6 +94,7 @@ type CertCache struct {
 	ocspFile     Updateable
 	ocspFilePath string
 	client       http.Client
+	developmentMode bool
 	// Domains to validate
 	Domains     []string
 	CertFile    string
@@ -114,7 +119,7 @@ type CertCache struct {
 // An alternative pattern would be to create an IsInitialized() bool or similarly named function that verifies all of the required fields have
 // been set. Then callers can just set fields in the struct by name and assert IsInitialized before doing anything with it.
 func New(certs []*x509.Certificate, certFetcher *certfetcher.CertFetcher, domains []string,
-	certFile string, newCertFile string, ocspCache string) *CertCache {
+	certFile string, newCertFile string, ocspCache string, developmentMode bool) *CertCache {
 	certName := ""
 	if len(certs) > 0 && certs[0] != nil {
 		certName = util.CertName(certs[0])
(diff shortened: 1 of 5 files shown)


twifkak:
Note to reviewer: Marking this as "draft" because I'd rather not add *another* merge conflict for you to deal with in #349, plus maybe you have a better suggestion than my sentinel value hack.

packager/certloader/certloader.go
```diff
@@ -41,7 +41,7 @@ func PopulateCertCache(config *util.Config, key crypto.PrivateKey, developmentMo
 			return nil, errors.Wrapf(err, "checking %s", config.CertFile)
 		}
 	}
-	certCache := certcache.New(certs, config.OCSPCache)
+	certCache := certcache.New(certs, config.OCSPCache, developmentMode)
```
banaag: I think you need to modify the certcache instantiation in gateway_server as well.

packager/certcache/certcache.go
```diff
@@ -51,6 +51,10 @@ const maxOCSPResponseBytes = 1024 * 1024
 // How often to check if OCSP stapling needs updating.
 const ocspCheckInterval = 1 * time.Hour
 
+// Sentinel value used to communicate that a returned OCSP was fake, and the caller should not attempt to parse it.
+// No, I'm not proud.
+var fakeOCSP = []byte("fake ocsp response")
```
shigeki: It is possible to create a valid OCSP response by using the private key of a self-signed cert such as gateway.go does. But it needs two certs to have the same private key.
https://github.com/ampproject/amppackager/blob/c9993b8ac4d17d1f05d3a1289956dadf3f9c370a/cmd/gateway_server/server.go#L148-L161

Or we can add an additional config to specify the issuer private key file to create a valid OCSP response.

twifkak: Thanks for the pointer; I didn't remember that gateway server did that. However, it seems that the gateway server doesn't fully exercise the certcache code, so it succeeds, whereas my attempt to adopt this fails, with "error validating fake response:  bad OCSP signature: x509: ECDSA verification failure". Can you take a look at https://play.golang.org/p/8owXw00LsRx and see if you can find the error in my code?

shigeki: The error can be resolved with the fix of  https://play.golang.org/p/CvVyUD-_tji . The private key needs to be the issuer key which corresponds to the self-signed cert key in this case.

twifkak: Ah, yes, that makes perfect sense. Thanks for the bug fix; I've updated this to generate a GOOD OCSP response as suggested.

PR #362, opened 2019-10-28
Add support for "PublicDir", PORT env var

ithinkihaveacat:
- if PublicDir specified, amppkg will look for flat files on filesystem in PublicDir directory
- use PORT env var for fallback port if not specified in config (needed for some hosts)


ithinkihaveacat:
Fixed the tests! @twifkak are you able to look at this, or let me know if you don't think these features are appropriate?

twifkak:
Thanks. I'm still catching up on reviews but this is on my backlog.

packager/signer/signer.go
```diff
@@ -138,13 +138,26 @@ func noRedirects(req *http.Request, via []*http.Request) error {
 	return http.ErrUseLastResponse
 }
 
-func New(certHandler certcache.CertHandler, key crypto.PrivateKey, urlSets []util.URLSet,
+func New(certHandler certcache.CertHandler, key crypto.PrivateKey, publicDir string, urlSets []util.URLSet,
 	rtvCache *rtv.RTVCache, shouldPackage func() error, overrideBaseURL *url.URL,
 	requireHeaders bool, forwardedRequestHeaders []string) (*Signer, error) {
+	var rt http.RoundTripper
```
twifkak: Ditto rt and t.

amppkg.example.toml
```diff
@@ -58,6 +58,10 @@ KeyFile = './pems/privkey.pem'
 # locking; consider this especially when utilizing network-mounted storage.
 OCSPCache = '/tmp/amppkg-ocsp'
 
+# If a PublicDir is specified, amppkg will look in this directory for
+# valid AMPs to package (instead of fetching AMPs from the network).
+# PublicDir = "public"
```
twifkak: Can you also explain what the path under that directory will be? Will it include the domain of the fetch URL or just its path?

Also, if it's just path -- is it possible to include the fetch domain in the disk path? It would be nice if this feature intersected well with the feature that you can have multiple URLSets.

packager/signer/signer.go
```diff
@@ -138,13 +138,26 @@ func noRedirects(req *http.Request, via []*http.Request) error {
...
+	if publicDir != "" {
+		t := &http.Transport{}
+		r := http.NewFileTransport(http.Dir(publicDir))
+		t.RegisterProtocol("http", r)
+		t.RegisterProtocol("https", r)
+		log.Printf("AMP source is \"%s\"", publicDir)
```
twifkak: Maybe s/AMP source is/AMP pages will be fetched from/ would be more understandable.

packager/signer/signer.go
```diff
@@ -138,13 +138,26 @@ func noRedirects(req *http.Request, via []*http.Request) error {
...
+		rt = t
+	} else {
+		rt = http.DefaultTransport
+		log.Printf("AMP source is network")
```
twifkak: Ditto.

packager/util/config.go
```diff
@@ -185,7 +187,11 @@ func ReadConfig(configBytes []byte) (*Config, error) {
 	// TODO(twifkak): Return an error if the TOML includes any fields that aren't part of the Config struct.
 
 	if config.Port == 0 {
-		config.Port = 8080
+		if i, err := strconv.ParseInt(os.Getenv("PORT"), 10, 32); err == nil {
```
twifkak: 1. Use [strconv.Atoi](https://golang.org/pkg/strconv/#Atoi) for a bit-depth that matches `int` so there won't be overflow.
2. If i <= 0, return an error.

twifkak: Can you document this somewhere? (amppkg.example.toml's Port field seems like the logical place.)

cmd/amppkg/main.go
```diff
@@ -147,7 +148,11 @@ func main() {
 
 	// TODO(twifkak): Add monitoring (e.g. per the above Cloudflare blog).
 
-	log.Println("Serving on port", config.Port)
+	var d []string
```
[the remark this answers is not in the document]
ithinkihaveacat: You and @sebastianbenz both…

packager/signer/signer.go
ithinkihaveacat: Yeah much better, thanks!

amppkg.example.toml
ithinkihaveacat: I've improved the comment.

I also had a look at including the fetch domain, but I can't see a good way to do this easily/cleanly... (Conceptually maybe the right place to put this is even in the `Scheme` key of `[URLSet.Fetch]`: if it's `file`, then we fetch from the local filesystem.)

But I hope it can be included as is! It's really only suitable for small test projects since it doesn't allow headers to be controlled, doesn't handle query strings, etc. (But, it does make these rather easier to deploy and reason about.)

ithinkihaveacat:
Updated! I think everything has been resolved save the feature request.

packager/signer/signer.go
```diff
@@ -138,13 +138,26 @@ func noRedirects(req *http.Request, via []*http.Request) error {
...
+func New(certHandler certcache.CertHandler, key crypto.PrivateKey, publicDir string, urlSets []util.URLSet,
 	rtvCache *rtv.RTVCache, shouldPackage func() error, overrideBaseURL *url.URL,
 	requireHeaders bool, forwardedRequestHeaders []string) (*Signer, error) {
+	var roundTripper http.RoundTripper
+	if publicDir != "" {
+		transport := &http.Transport{}
+		r := http.NewFileTransport(http.Dir(publicDir))
+		transport.RegisterProtocol("http", r)
+		transport.RegisterProtocol("https", r)
+		log.Printf("AMP pages will be fetched from the directory \"%s\"", publicDir)
```
twifkak: Ditto \n.

packager/signer/signer.go
```diff
@@ -138,13 +138,26 @@ func noRedirects(req *http.Request, via []*http.Request) error {
...
+		roundTripper = transport
+	} else {
+		roundTripper = http.DefaultTransport
+		log.Printf("AMP pages will be fetched from network")
```
twifkak: Ditto \n.

packager/util/config.go
```diff
@@ -200,7 +202,15 @@ func ReadConfig(configBytes []byte) (*Config, error) {
 	// TODO(twifkak): Return an error if the TOML includes any fields that aren't part of the Config struct.
 
 	if config.Port == 0 {
-		config.Port = 8080
+		if os.Getenv("PORT") != "" {
```
twifkak: (nit) Can you make a local var for `os.Getenv("PORT")` so you don't have to call it again on lines 206 and 209? Not an actual perf or correctness thing, just an aesthetic thing.

packager/signer/signer_test.go
```diff
@@ -78,7 +78,7 @@ type SignerSuite struct {
 
 func (this *SignerSuite) new(urlSets []util.URLSet) http.Handler {
 	forwardedRequestHeaders := []string{"Host", "X-Foo"}
-	handler, err := New(fakeCertHandler{}, pkgt.Key, urlSets, &rtv.RTVCache{}, func() error { return this.shouldPackage }, nil, true, forwardedRequestHeaders)
+	handler, err := New(fakeCertHandler{}, pkgt.Key, "", urlSets, &rtv.RTVCache{}, func() error { return this.shouldPackage }, nil, true, forwardedRequestHeaders)
 	this.Require().NoError(err)
```
twifkak: Optionally, a test for this functionality would be cool (make a new subdir under testdata with an html file, I guess). But no big deal if it turns out to be too ornery to set up.

cmd/amppkg/main.go
```diff
@@ -151,7 +152,11 @@ func main() {
 
 	// TODO(twifkak): Add monitoring (e.g. per the above Cloudflare blog).
 
-	log.Println("Serving on port", config.Port)
+	var domains []string
+	for _, urlset := range config.URLSet {
+		domains = append(domains, urlset.Sign.Domain)
+	}
+	log.Printf("Serving SXG for \"%s\" on port %d\n", strings.Join(domains, ", "), config.Port)
```
twifkak: This is missing a \n at the end.

amppkg.example.toml
twifkak: OK, that's fine; can be included as-is. I guess it might be possible if you implemented a custom RoundTripper that delegated to FileTransport with a modified path? Dunno, not a big deal either way.

twifkak:
Hi @ithinkihaveacat - Are you still interested in this? Do you still have time to work on it? Want to know whether I should leave it be, close without merging, or take over the work.

ithinkihaveacat:
Sorry for the big delay! I still want to get this merged. Will update in the next week.

ithinkihaveacat:
Sorry to leave this hanging. Had a few goes at trying to complete this, but am not really making much progress. Will close it for now!

PR #367, opened 2019-11-13
Multi domain cert/key support

alabiaga:


alabiaga:
closing..local files in incorrect golang project structure.

PR #379, opened 2019-12-19
make stop channel work properly for both maintainCerts and maintainOCSP

tomokinat:
Current implementation passes `stop` channel to both `maintainCerts` and `maintainOCSP` functions. It won't work as expected unless `stop` is buffered (e.g. `make(chan struct{}, 2)`) and the value is sent twice, which will result in a arcane parameter precondition of an exported function `CertCache.Init`.

This PR fixes that problem in another way. [PoC](https://play.golang.org/p/8MVG7ZdqDCq)

### DIFF
--- a/cmd/amppkg/main.go
+++ b/cmd/amppkg/main.go
@@ -98,7 +98,7 @@ func main() {
 		die(errors.Wrap(err, "building cert cache"))
 	}
 
-	if err = certCache.Init(nil); err != nil {
+	if err = certCache.Init(); err != nil {
 		if *flagDevelopment {
 			fmt.Println("WARNING:", err)
 		} else {
--- a/packager/certcache/certcache.go
+++ b/packager/certcache/certcache.go
@@ -88,6 +88,7 @@ type CertCache struct {
 	renewedCerts      []*x509.Certificate
 	ocspUpdateAfterMu sync.RWMutex
 	ocspUpdateAfter   time.Time
+	stop              chan struct{}
 	// TODO(twifkak): Implement a registry of Updateable instances which can be configured in the toml.
 	ocspFile     Updateable
 	ocspFilePath string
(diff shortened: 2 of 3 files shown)


tomokinat:
CI's failure seems due to this flaky test :(
https://github.com/tomokinat/amppackager/blob/fa4e932ef896aab256f89a155bf150928b2dbd85/packager/certcache/certcache_test.go#L227

twifkak:
Good find! I think a simpler solution may be possible, per https://blog.golang.org/pipelines#TOC_6%2e:

> We need a way to tell an unknown and unbounded number of goroutines to stop sending their values downstream. In Go, we can do this by closing a channel, because a receive operation on a closed channel can always proceed immediately, yielding the element type's zero value.

So can you try changing the code in main() to close the channel instead of sending a value?

If that doesn't work, then please extract this channel-forking code into a separate function and leave a comment explaining it.

Sorry about the flake! Restarted the build.

tomokinat:
Yes it works :) Thanks for the suggestion.

Yet I still feel like "pass to me `stop chan struct{}` and close it if you want to stop me" isn't a good I/F.
How about returning a function, similar to https://golang.org/pkg/context/#WithCancel ?

[code](https://play.golang.org/p/b2Dkrz05jnk)

twifkak:
I'm okay changing the interface. Rather than returning a func from Init(), what about adding `func (this *CertCache) Stop()`? Init() could store the channel(s) in the CertCache struct.

tomokinat:
I just realized that if `CertCache.Init` is assumed to be callable even after once stopped for re-starting, `stop` should be initialized in `Init`, not in `New`. Should it be?

twifkak:
No, I think we only designed `Init` to be called once, and to be the first call after `New`. Should I merge now, or do you want to make any other changes?

tomokinat:
No, now this should be ready to merge :)

twifkak:
Thanks! (On second thought, I guess `make` would make sense inside `New`, but it's no big deal either way.)

PR #380, opened 2019-12-19
Bump dependency on github.com/go-acme/lego to v3.2.0 to get rid of indirect dependency on a now-unknown code commit

gabbifish:
This PR bumps github.com/go-acme/lego to v3.2.0 to fix the indirect dependency on a now-unavailable revision:
```
go: finding github.com/labbsr0x/goh v0.0.0-20190417202808-8b16b4848295
go: github.com/labbsr0x/goh@v0.0.0-20190417202808-8b16b4848295: unknown revision 8b16b4848295
```

cc @twifkak


gabbifish:
@googlebot I signed it!

twifkak:
I have more to learn about go modules. Why is the original one failing? I see that commit hash [`8b16b4848295`](https://github.com/labbsr0x/goh/commit/8b16b4848295) still exists, but I don't know if that means the go version `v0.0.0-20190417202808-8b16b4848295` still does.

When I ran `go get -v -x github.com/labbsr0x/goh@v0.0.0-20190417202808-8b16b4848295` from a different go package, I got these errors:
```
# get https://proxy.golang.org/github.com/@v/v0.0.0-20190417202808-8b16b4848295.info: 410 Gone (1.479s)
# get https://proxy.golang.org/github.com/labbsr0x/@v/v0.0.0-20190417202808-8b16b4848295.info: 410 Gone (1.623s)
```

And those URLs turn up even weirder errors:

```
not found: github.com@v0.0.0-20190417202808-8b16b4848295: unrecognized import path "github.com" (parse https://github.com/?go-get=1: no go-import meta tags ())
not found: github.com/labbsr0x@v0.0.0-20190417202808-8b16b4848295: invalid github.com/ import path "github.com/labbsr0x"
```

I'd thought it might have to do with the `v0.0.0` label, but it seems that's some sort of sentinel "don't look for a tag" label, given that e.g. `github.com/golang/snappy` is getting along fine with a `v0.0.0` label even though it doesn't have such a tag.

Anyway, understanding all that may be a problem for another day.

gabbifish:
@twifkak this also really confuses me. If this were a direct dependency, I would have `go get`-ted the missing dependency at that exact hash, but since this is an indirect dependency, I can't really do that, I think. 

I fumbled around trying to find a satisfactory answer for a bit, but so far I haven't found one...

twifkak:
Oh, BTW, thanks for fixing this!

PR #382, opened 2020-01-17
Add Forwarded headers to the fetch request.

### LINKED ISSUE #260: Include limited `Forwarded` in fetch request
Setting the [`Forwarded` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) in its entirety would increase the risk of [over-signing](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#seccons-over-signing); however, it seems like only the `for` directive of that header is a concern; the others (`by`, `host`, and `proto`) seem okay in all but the most unusual cases [1]. Perhaps just `host` and `proto` since those are safest and what's really needed by most backend servers.

Consider also sending the legacy headers [`X-Forwarded-Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) and [`X-Forwarded-Proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto).

[Previous discussion](https://github.com/ampproject/amppackager/pull/256#issuecomment-471048404).

[1] Wild hypothesizing here: perhaps some software-defined networking thing that spawns a new virtual network interface for each session.

twifkak:
To be cautious re: privacy/security, specify only host. Note that this
changes behavior for anybody who was previously including "Forwarded" or
"X-Forwarded-Host" in their ForwardedRequestHeaders in the config.

Fixes #260.

### DIFF
--- a/packager/signer/signer_test.go
+++ b/packager/signer/signer_test.go
@@ -183,6 +183,8 @@ func (this *SignerSuite) TestSimple() {
 	this.Assert().Equal(fakePath, this.lastRequest.URL.String())
 	this.Assert().Equal(userAgent, this.lastRequest.Header.Get("User-Agent"))
 	this.Assert().Equal("1.1 amppkg", this.lastRequest.Header.Get("Via"))
+	this.Assert().Equal(`host="example.com"`, this.lastRequest.Header.Get("Forwarded"))
+	this.Assert().Equal("example.com", this.lastRequest.Header.Get("X-Forwarded-Host"))
 	this.Assert().Equal(http.StatusOK, resp.StatusCode, "incorrect status: %#v", resp)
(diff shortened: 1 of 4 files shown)


packager/signer/signer_test.go
```diff
@@ -183,6 +183,8 @@ func (this *SignerSuite) TestSimple() {
 	this.Assert().Equal(fakePath, this.lastRequest.URL.String())
 	this.Assert().Equal(userAgent, this.lastRequest.Header.Get("User-Agent"))
 	this.Assert().Equal("1.1 amppkg", this.lastRequest.Header.Get("Via"))
+	this.Assert().Equal(`host="example.com"`, this.lastRequest.Header.Get("Forwarded"))
```
banaag: Could you add a test that has other parameters apart from host?

Issue #395, opened 2020-02-19
Leaving out max-age on amp-script leads to invalid signature


dritter:
Hi there!

We use an `amp-script` with `script` attribute, but without specifying a `max-age`. When trying to verify the signature, it turns out to be invalid (message: "sig_expires is in the past").. AFAIK the default expiration time for that case is 1 day.

Looking over the code, it looks like the expiration is calculated based on [24 hours prior to now](https://github.com/ampproject/amppackager/blame/11c1e4c3ea993b0569310b82edcd4e1332494788/packager/signer/signer.go#L497). That is rather odd to me. Wouldn't that code lead to an expiration of 0 (aka the current time)? And secondly, isn't the minimal max-age > 86400 now?

We worked around that by setting a max-age to a high value (1 year) explicitly.

Thanks in advance.

twifkak:
Hi, thanks for the report. I agree, serving an expired SXG is odd. It's a confluence of two deliberate choices. I think the thing to do in this case is just not sign at all. I don't think it's a high priority change, as it won't have an impact on behavior in the Google AMP Cache, which already treats this case as unsigned. Let me know if you disagree. Details:

#### Clock skew
We want to serve SXGs that are somewhat forgiving of client-side clock skew. Section 7.1 of [this paper on cert errors](https://acmccs.github.io/papers/p1407-acerA.pdf) suggests that client clocks are much more often behind than ahead, and says that 93.3% of HTTPS requests have a skew of less than 24h. It doesn't have more fine-grained details within that 24h window. I'd love to see more public data/analysis on this subject.

#### Downgrades
We want to minimize the security considerations of enabling AMP Packager, and of writing an inline `<amp-script>`. There is a potential spooky-action-at-a-distance in that the interaction of SXG and inline JS requires some thought. But in reality, the publisher choice to enable SXG and choice to write some inline JS are probably made independently and without consideration for one another. Forcing the script-writer to specify a max-age eliminates that spooky-action-at-a-distance.

One example scenario is: 1) publisher has a vulnerability in their JS, 2) attacker downloads an SXG of that vulnerability, 3) publisher fixes it, 4) attacker serves the SXG to people. Thus, the lifetime of the vulnerability is extended to the lifetime of its SXG. 24h was chosen as a default max-age because that's the [default max-age of service workers](https://chromium.googlesource.com/chromium/src/+/master/docs/security/service-worker-security-faq.md#do-service-workers-live-forever), which have a similar lifetime-extension property.

[...]

PR #396, opened 2020-02-20
Proxy unsigned if expires is in the past.

twifkak:
For pages with amp-script max-age<=86400, this will return plain HTML
rather than generating an already-expired SXG.

Addresses #395.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -503,7 +503,7 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 	}
 	signer := signedexchange.Signer{
 		Date:        date,
-		Expires:     date.Add(duration),
+		Expires:     expires,
 		Certs:       []*x509.Certificate{cert},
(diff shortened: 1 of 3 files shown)


twifkak:
cc @dritter

packager/signer/signer.go
```diff
@@ -495,6 +495,12 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
...
 	date := now.Add(-24 * time.Hour)
+	expires := date.Add(duration)
+	if !expires.After(now) {
+		log.Printf("Not packaging because computed max-age %d places expiry in the past\n", metadata.MaxAgeSecs)
+		proxy(resp, fetchResp, fetchBody)
+		return
+	}
 	signer := signedexchange.Signer{
 		Date:        date,
 		Expires:     date.Add(duration),
```
banaag: Should you use "expires" (from above) here?

twifkak: Oops, yes.

PR #399, opened 2020-02-27
Do not log an OCSP error in case of cert renewal

shigeki:
An old ocsp cache causes a parse error of
```
Error computing OCSP midpoint: Parsing OCSP: no response matching the supplied certificate
```
when the cert is renewed and the process is restarted. 
Do not log it in its initialization.


banaag:
LGTM but wait for feedback from twifkak@ as well.

PR #402, opened 2020-03-05
Updated Dockerfiles for consumer and renewer

banaag:
Updated Dockerfiles for consumer and renewer to use multi-stage and reduce docker image sizes from 2GB to 12.3MB

### DIFF
--- a/deploy/gcloud/Dockerfile.renewer
+++ b/deploy/gcloud/Dockerfile.renewer
@@ -12,25 +12,60 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Used https://medium.com/@chemidy/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324
+# as a guide.
+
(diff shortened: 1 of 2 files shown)


deploy/gcloud/Dockerfile.renewer
```diff
@@ -12,25 +12,60 @@
...
 EXPOSE 8080
 
+ENV PATH=$PATH:.
+
 # Start the AMP Packager
 ENTRYPOINT ["amppkg"]
 
-# Set default flags to run in development mode.
+# Set default flags to run in autorenew cert mode.
 CMD ["-autorenewcert", "-config=/renewer/amppkg_renewer.toml"]
```
twifkak: Have you considered making this inherit from Dockerfile.consumer to reduce duplication?

banaag: Certainly possible but decided to keep them separate for now until everything settles down. We may revisit for a future PR.

Issue #357, opened 2019-10-21
Prometheus endpoint for metrics


FranGM:
Hey there!

I think a prometheus endpoint would be very handy to have in this service, so we can easily track request/error rates etc without needing to rely on the metrics for the proxied service itself. Once the service is out of development mode it's not easy to operate without having metrics readily available.

Thanks!

twifkak:
Oops, I missed this notification. Agreed, seems like a good idea. Any suggestions for metrics to expose? Off the top of my head:
 - # of requests
 - failures (by type)

Less important but possibly useful:
 - per-request: CPU usage, response byte size, latency (total [request -> response], added [fetch response -> response])

FranGM:
Yeah I think those are pretty much what I had in mind.

Response byte size is definitely important, given that there's a limit on the size of the response that the packager can process, so being able to monitor that would be really useful.

The ability to monitor response codes from upstream as well as latency would also be great.

Thanks for working on this!

twifkak:
Sorry for the delay on this. It's still top-of-mind, but life.

Wanted to document a few other ideas:
 - time until cert expires
 - time until ocsp expires
 - ACME or OCSP fetch errors

MichaelRybak:
@FranGM Hi Fran!

I've implemented a bunch of basic metrics recently based per your request. The [documentation](https://github.com/ampproject/amppackager/blob/master/monitoring.md) lists the available metrics. Would you like to try them out and give us some feedback?

FranGM:
Hey there! Sorry about the delay. I might not have enough time to test this in any meaningful way at the moment but I've been going over the linked docs (which are excellent) and this seems exactly what we need. Thanks!

MichaelRybak:
Happy to hear that!

PR #413, opened 2020-04-28
Add mux tests.

MichaelRybak:
Will allow safely adding a Prometheus metrics endpoint: #357.

### DIFF
--- a/packager/mux/mux_test.go
+++ b/packager/mux/mux_test.go
@@ -1,0 +1,219 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mux
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	pkgt "github.com/ampproject/amppackager/packager/testing"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+// expand propagates hardcoded values into template test url.
+func expand(templateURL string) string {
+	templateURL = strings.Replace(templateURL, "$HOST", "http://www.publisher_amp_server.com", 1)
+	templateURL = strings.Replace(templateURL, "$FETCH", "http://www.publisher_main_server.com/some_page", 1)
+	templateURL = strings.Replace(templateURL, "$SIGN", "https://www.publisher_main_server.com/some_page", 1)
+	templateURL = strings.Replace(templateURL, "$CERT", pkgt.CertName, 1)
+	return templateURL
+}
+
+// mockedHandler mocks mux' underlying http handlers - signer, cert etc.
+type mockedHandler struct {
+	mock.Mock
+}
+
+func (m *mockedHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	m.Called(Params(req))
+}
+
+func TestServeHTTPSuccess(t *testing.T) {
+	templateTests := []struct {
+		testName      string
+		testURL       string
+		expectHandler string
+		expectParams  map[string]string
+	}{
+		{
+			testName:      `Signer - empty`,
+			testURL:       `$HOST/priv/doc`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{},
+		}, {
+			testName:      `Signer - with query, empty`,
+			testURL:       `$HOST/priv/doc?`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{},
+		}, {
+			testName:      `Signer - with query, regular`,
+			testURL:       `$HOST/priv/doc?fetch=$FETCH&sign=$SIGN`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{},
+		},
+		{
+			testName:      `Signer - with query, escaping`,
+			testURL:       `$HOST/priv/doc?fetch=$FETCH&sign=$SIGN%2A\`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{},
+		}, {
+			testName:      `Signer - with path, empty`,
+			testURL:       `$HOST/priv/doc/`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{`signURL`: ``},
+		}, {
+			testName:      `Signer -  with path, regular`,
+			testURL:       `$HOST/priv/doc/$FETCH`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{`signURL`: `$FETCH`},
+		}, {
+			testName:      `Signer - with path, escaping`,
+			testURL:       `$HOST/priv/doc/$FETCH%2A\`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{`signURL`: `$FETCH%2A%5C`},
+		}, {
+			testName:      `Signer - with path and query, regular`,
+			testURL:       `$HOST/priv/doc/$FETCH?amp=1`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{`signURL`: `$FETCH?amp=1`},
+		}, {
+			testName:      `Signer - with path and query, escaping`,
+			testURL:       `$HOST/priv/doc/$FETCH%2A\?amp=1%2A\`,
+			expectHandler: `signer`,
+			expectParams:  map[string]string{`signURL`: `$FETCH%2A%5C?amp=1%2A\`},
+		}, {
+			testName:      `Cert - empty`,
+			testURL:       `$HOST/amppkg/cert/`,
+			expectHandler: `cert`,
+			expectParams:  map[string]string{`certName`: ``},
+		}, {
+			testName:      `Cert - regular`,
+			testURL:       `$HOST/amppkg/cert/$CERT`,
+			expectHandler: `cert`,
+			expectParams:  map[string]string{`certName`: `$CERT`},
+		}, {
+			testName:      `Cert - escaping`,
+			testURL:       `$HOST/amppkg/cert/$CERT%2A\`,
+			expectHandler: `cert`,
+			expectParams:  map[string]string{`certName`: `$CERT*\`},
+		}, {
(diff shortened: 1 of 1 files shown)


banaag:
Looks like travis is failing: https://travis-ci.org/github/ampproject/amppackager/builds/680776394
Could you take a look? Thanks.

MichaelRybak:
> Looks like travis is failing: https://travis-ci.org/github/ampproject/amppackager/builds/680776394
> Could you take a look? Thanks.

Thanks, fixed.

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"testing"
+
+	pkgt "github.com/ampproject/amppackager/packager/testing"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+// Propagate hardcoded values into template test url.
+func Expand(templateURL string) string {
+	templateURL = strings.Replace(templateURL, "$HOST",
```
banaag, line 34: Avoid line breaks during function calls. Per: go/gocomments#line-length
Ditto for other lines below.

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	pkgt "github.com/ampproject/amppackager/packager/testing"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+// Propagate hardcoded values into template test url.
```
banaag, line 32: Comments should be full sentences as per:  go/gocomments#comment-sentences
Expand propagates hardcoded values into ...
Similarly for other items below.

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+		"http://www.publisher_amp_server.com", 1)
+	templateURL = strings.Replace(templateURL, "$FETCH",
+		"http://www.publisher_main_server.com/some_page", 1)
+	templateURL = strings.Replace(templateURL, "$SIGN",
+		"https://www.publisher_main_server.com/some_page", 1)
+	templateURL = strings.Replace(templateURL, "$CERT", pkgt.CertName, 1)
+	return templateURL
+}
+
+// Dedicated type for annotated params - for tests readability.
```
banaag, line 44: go/gocomments#comment-sentences

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+type params map[string]string
+
+// Mock for underlying http handlers - signer, cert etc.
```
banaag, line 47: go/gocomments#comment-sentences

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+func TestServeHTTPSuccess(t *testing.T) {
+	tests := []struct {
+		expectedHandlerName  string
+		testScenario         string
+		url                  string
+		expectedParsedParams params
+	}{
+		{"signer",
+			" empty",
```
banaag, line 66: Something odd about indentation here and below.  Try gofmt or goimports?

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+			// Set expectation.
+			expectedMockedHandler := mocks[tt.expectedHandlerName]
+			expectedMockedHandler.On("ServeHTTP", tt.expectedParsedParams)
+
+			// Run.
+			actualResp := pkgt.Get(t, mux, tt.url)
+
+			// Expect no errors.
+			assert.Equal(t, 200, actualResp.StatusCode, "No error expected: %#v", actualResp)
```
banaag, line 136: optional: I learned this when my changes for another project got reviewed by another team that we shouldn't be using assert in unit tests for go, so it's ok if you want to keep it this way.
go/gocomments#assert

packager/mux/mux_test.go
```diff
@@ -0,0 +1,199 @@
...
+	mockedHandler := new(MockedHandler)
+	mux := New(mockedHandler, mockedHandler, mockedHandler, mockedHandler)
+
+	// Run and extract error.
+	actualResp := pkgt.GetBHH(t, mux, url, "", body, http.Header{})
+	actualErrorMessageBuffer, _ := ioutil.ReadAll(actualResp.Body)
+	actualErrorMessage := fmt.Sprintf("%s", actualErrorMessageBuffer)
+
+	// Expect the right error.
+	assert.Equal(t, expectedErrorCode, actualResp.StatusCode,
```
banaag, line 157: go/go-test-comments#identify-the-function

packager/mux/mux_test.go
MichaelRybak: Done.

packager/mux/mux_test.go
MichaelRybak: Done. I also de-capitalized internal functions (after learning that capitalization exports them).

packager/mux/mux_test.go
MichaelRybak: After re-formatting the tests I felt like this dedicated type was no longer helping readability and removed it.

packager/mux/mux_test.go
MichaelRybak: Done.

packager/mux/mux_test.go
MichaelRybak: Done. Also switched to field:value initializations for readability.

packager/mux/mux_test.go
MichaelRybak: Hm, I've looked into this, and also the relevant [FAQ](https://golang.org/doc/faq#assertions), and looks like asserts are actually more helpful than the go's native way.

My thoughts on go's FAQ:

> Go doesn't provide assertions. They are undeniably convenient, but our experience has been that programmers use them as a crutch to avoid thinking about proper error handling and reporting. Proper error handling means that servers continue to operate instead of crashing after a non-fatal error.

For tests it's actually helpful to terminate the test automatically on first failed assertion. E.g. if a functional test parses a URL from XML and then fetches it, I'd like the test to stop if parsing failed - instead of trying to fetch a broken URL. So assertion is kind of like throwing an unhandled exception, with is acceptable and convenient for tests IMO.

> Proper error reporting means that errors are direct and to the point, saving the programmer from interpreting a large crash trace. Precise errors are particularly important when the programmer seeing the errors is not familiar with the code.

I agree. Yet golang's native way requires more work to implement error reporting than testify/assert package. 

testify/assert provides precise reporting for free. When an assertion fails, the framework provides the test name, the type of musmatch (e.g. "not equal"), the mismatched values, the diff, and the line number:

```
--- FAIL: TestParamsIncorrectValueType (0.00s)
	mux_test.go:209: 
			Error Trace:	mux_test.go:209
			Error:      	Not equal: 
			            	expected: map[string]string{}
			            	actual  : map[string]string{"":""}
			            	
			            	Diff:
			            	--- Expected
			            	+++ Actual
			            	@@ -1,2 +1,3 @@
			            	-(map[string]string) {
			            	+(map[string]string) (len=1) {
			            	+ (string) "": (string) ""
			            	 }
			Test:       	TestParamsIncorrectValueType
FAIL
```

On the contrary, the native "t.Error" output looks like this:

```
--- FAIL: TestParamsIncorrectValueType (0.00s)
	mux_test.go:210: 
FAIL
```

packager/mux/mux_test.go
MichaelRybak: I've removed the custom error altogether, since testify asserts provide verbose output already:

```
-- FAIL: TestServeHTTPexpect404s (0.00s)
    --- FAIL: TestServeHTTPexpect404s/No_such_endpoint______________________ (0.00s)
    	mux_test.go:171: 
    			Error Trace:	mux_test.go:171
    			            				mux_test.go:192
    			Error:      	Not equal: 
    			            	expected: 405
    			            	actual  : 404
    			Test:       	TestServeHTTPexpect404s/No_such_endpoint______________________
    --- FAIL: TestServeHTTPexpect404s/Signer_-_unexpected_extra_char________ (0.00s)
    	mux_test.go:171: 
    			Error Trace:	mux_test.go:171
    			            				mux_test.go:192
    			Error:      	Not equal: 
    			            	expected: 405
    			            	actual  : 404
    			Test:       	TestServeHTTPexpect404s/Signer_-_unexpected_extra_char________
```

MichaelRybak:
> Thanks for the fixes. LGTM but you should wait for twifkak@ to comment/approve.

Thanks for review! I've now replied to all your comments, including the one about using asserts. Letting you know because I was still typing for a while after updating the PR :)

packager/mux/mux_test.go
banaag: I am also on the side of using asserts, that is why I marked this comment as optional. I am actually not sure why the go style guide promotes non-use of asserts but I'm sure it's not without controversy.

twifkak: I'm also on the side of keeping the asserts! Had this comment sitting in draft:

I was unconvinced by the reasoning behind that go/ link (and since this codebase lives outside of Google, I don't think we're beholden to that doc). My reasoning for using the assert lib everywhere is that it improves readability, by removing boilerplate and letting the reader focus on business logic.

packager/mux/mux_test.go
```diff
@@ -0,0 +1,210 @@
...
+			expectParams:  map[string]string{},
+		}, {
+			testName:      "Signer - with query, regular",
+			testURL:       "$HOST/priv/doc?fetch=$FETCH&sign=$SIGN",
+			expectHandler: "signer",
+			expectParams:  map[string]string{},
+		},
+		{
+			testName:      "Signer - with query, escaping",
+			testURL:       "$HOST/priv/doc?fetch=$FETCH&sign=$SIGN%2A\\",
```
twifkak, line 75: Optionally, use the \`backtick\` string literal syntax to avoid the double backslash.

packager/mux/mux_test.go
```diff
@@ -0,0 +1,210 @@
...
+			expectMockedHandler.On("ServeHTTP", tt.expectParams)
+
+			// Run.
+			actualResp := pkgt.Get(t, mux, tt.testURL)
+
+			// Expect no errors.
+			assert.Equal(t, 200, actualResp.StatusCode, "No error expected: %#v", actualResp)
+
+			// Expect the right call to the right handler, and no calls to the rest.
+			for _, mockedHandler := range mocks {
```
twifkak, line 153: Optionally, put this right after you define the mocks, using defer. e.g.

```
defer func() {
  for _, mockedHandler ... {
    ...
  }
}()
```

I generally like this pattern for cleanup because it deals gracefully with early returns and panics. (Similar to `finally`, `__exit__`, and RAII from other languages.)

In this particular case it doesn't make much of a difference because you have no early returns, but it still provides some peace of mind right up front that the cleanup will be handled later.

packager/mux/mux_test.go
MichaelRybak: Will do.

packager/mux/mux_test.go
MichaelRybak: Cool, will do!

PR #414, opened 2020-04-30
Update README.

MichaelRybak:


README.md
```diff
@@ -31,13 +31,18 @@ own and can obtain certificates for.
 
 ##### Manual installation
 
-  1. Install Go version 1.10 or higher. Optionally, set
+  1. Install Go version 1.10. Optionally, set
```
MichaelRybak, line 34: Google will validate all contributions against Travis running Go 1.10, so should we suggest installing specifically this version?

I've installed Go 1.14, used strings.ReplaceAll and got a build error from Travis:
https://travis-ci.org/github/ampproject/amppackager/builds/680776394

README.md
```diff
@@ -31,13 +31,18 @@ own and can obtain certificates for.
...
      [$GOPATH](https://github.com/golang/go/wiki/GOPATH) to something (default
      is `~/go`) and/or add `$GOPATH/bin` to `$PATH`.
-  2. `go get -u -mod=vendor github.com/ampproject/amppackager/cmd/amppkg`
+  1. `go get -u github.com/ampproject/amppackager/cmd/amppkg`
```
MichaelRybak, line 37: -mod=vendor is not available in Go 1.14: https://github.com/GoogleContainerTools/skaffold/issues/3699

I've tested it today with Go 1.10 and it didn't work for me either (but it did work when I removed -mod=vendor):

michaelrybak$ go get -u -mod=vendor github.com/ampproject/amppackager/cmd/amppkg
flag provided but not defined: -mod
usage: get [-d] [-f] [-fix] [-insecure] [-t] [-u] [-v] [build flags] [packages]
Run 'go help get' for details.

README.md
```diff
@@ -63,10 +68,15 @@ container.
 
   1. Run Chrome with the following commandline flags:
      ```
-     --user-data-dir=/tmp/udd
-     --ignore-certificate-errors-spki-list=$(openssl x509 -pubkey -noout -in path/to/fullchain.pem | openssl pkey -pubin -outform der | openssl dgst -sha256 -binary | base64)
-     --enable-features=SignedHTTPExchange
-     'data:text/html,<a href="https://localhost:8080/priv/doc/https://amppackageexample.com/">click me'
+     alias chrome = [FULL PATH TO CHROME BINARY]
+     PATH_TO_FULLCHAIN_PEM = [FULL PATH TO fullchain.pem]
```
MichaelRybak, line 72: I didn't notice "path/to/fullchain.pem" in the command. It will be easier to notice this way.

README.md
```diff
@@ -63,10 +68,15 @@ container.
...
+     chrome --user-data-dir=/tmp/udd\
+         --ignore-certificate-errors-spki-list=$(\
```
MichaelRybak, line 74: Added line breaks for readability.

README.md
banaag: @twifkak perhaps it's time to update travis.yml? What do you think?

README.md
banaag: @twifkak, do we still need to generate the vendor directory?

banaag:
The changes look reasonable to me but you should wait for twifkak@ to bless.

README.md
twifkak: I wouldn't remove the "or higher" unless there's a known breakage at higher versions. Don't want to make installation unnecessarily hard. I guess I have to exclude `-mod=vendor` from that rule. :)

Re: updating travis.yml, SGTM. It may even be possible to test on multiple versions? e.g. https://github.com/streadway/amqp/blob/master/.travis.yml

README.md
twifkak: Oof. They're not making it easy. I guess we should list two command lines: with `-mod=vendor` if 1.10-1.13, and without if 1.14+. I think it's too soon to up our minimum version to 1.14 -- e.g. it's not even in the Ubuntu release from 7 days ago: https://packages.ubuntu.com/focal/golang-any, let alone in the previous LTS version.

We could just do as you've suggested, which AFAIK means that the vendor directory will be ignored on Go 1.10-1.13. I'm not wild about that, for two reasons:
 1. I think it's slow? Involves a lot of `git clone`s.
 2. I'm just not knowledgeable enough in how go modules work. After doing all those `git clone`s, does Go verify against `go.sum`? I would prefer some safety check that users are getting the version of the libs that we tested against.

I could easily be convinced on both fronts.

(To make things more confusing, it seems the official documentation is incorrect. https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away and https://golang.org/doc/go1.14#go-command seem to suggest continued support for `-mod=vendor`. But the commit that you link to is unequivocal.)

README.md
twifkak: LGTM

README.md
twifkak: LGTM

MichaelRybak:
@twifkak @banaag Devin, Allan, I've removed "mod=vendor" and "1.10 or higher" updates - could you LGTM what's left? We can have this PR pulled in, and update the version/modules stuff later.

README.md
```diff
@@ -34,10 +34,15 @@ own and can obtain certificates for.
   1. Install Go version 1.10 or higher. Optionally, set
      [$GOPATH](https://github.com/golang/go/wiki/GOPATH) to something (default
      is `~/go`) and/or add `$GOPATH/bin` to `$PATH`.
-  2. `go get -u -mod=vendor github.com/ampproject/amppackager/cmd/amppkg`
+  1. `go get -u -mod=vendor github.com/ampproject/amppackager/cmd/amppkg`
```
twifkak, line 37: (optional) For now, I'd be OK with putting an if statement here based on `go version`, until we work out the details.

MichaelRybak: Done.

PR #416, opened 2020-05-04
Refactor mux - introduce the routing matrix.

MichaelRybak:
Refactor mux. Will simplify adding Prometheus metrics endpoint (#357).

### DIFF
--- a/packager/mux/mux.go
+++ b/packager/mux/mux.go
@@ -41,35 +41,92 @@ import (
 	"github.com/ampproject/amppackager/packager/util"
 )
 
+// routingRule maps a URL path prefix to three entities:
+// * suffixValidatorFunc - a function that validates the suffix of URL path,
+// * handler - an http.Handler that should handle such prefix,
+// * handlerPrometheusLabel - a label (dimension) to be used in
+//       handler-agnostic Prometheus metrics like requests count.
+//       Must adhere to the Prometheus data model:
+// 		 https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+type routingRule struct {
+	urlPathPrefix          string
+	suffixValidatorFunc    func(suffix string, req *http.Request, params *map[string]string, errorMsg *string, errorCode *int)
+	handler                http.Handler
+	handlerPrometheusLabel string
+}
+
+// mux stores a routingMatrix, an array of routing rules that define the mux'
+// routing logic. Note that the order of rules in the matrix is important: the
+// first matching rule will be applied by ServeHTTP, so the rule for a
+// particular prefix should precede the rule for it's sub-prefix. E.g. the rule
+// for routing requests prefixed with “/priv/doc/” (w/ trailing slash) must go
+// before the rule for routing requests prefixed with “/priv/doc” (w/o trailing
+// slash).
+// The last rule in the matrix must have empty urlPathPrefix, so it would match
+// any URL. This ensures there's at least one matching rule for any URL.
 type mux struct {
-	certCache   http.Handler
-	signer      http.Handler
-	validityMap http.Handler
-	healthz     http.Handler
+	routingMatrix []routingRule
 }
 
-// The main entry point. Use the return value for http.Server.Handler.
+// return404 is a URL Path Suffix Validator that always returns 404.
+func return404(suffix string, req *http.Request, params *map[string]string, errorMsg *string, errorCode *int) {
+	*errorMsg, *errorCode = "404 page not found", http.StatusNotFound
+}
+
+// expectNoSuffix is a URL Path Suffix Validator that expects an empty suffix.
+func expectNoSuffix(suffix string, req *http.Request, params *map[string]string, errorMsg *string, errorCode *int) {
+	if suffix != "" {
+		return404(suffix, req, params, errorMsg, errorCode)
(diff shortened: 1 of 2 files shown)


packager/mux/mux.go
```diff
@@ -37,82 +37,134 @@ import (
...
-	params := map[string]string{}
-	req = WithParams(req, params)
-
-	// Use EscapedPath rather than RequestURI because the latter can take
-	// absolute-form, per https://tools.ietf.org/html/rfc7230#section-5.3.
-	//
-	// Use EscapedPath rather than RawPath to verify that the request URI
-	// is a valid encoding, per
-	// https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange
-	// item 3.
```
twifkak, line 79: I'd prefer to keep this explanation of why EscapedPath() was chosen over the alternatives, because it's not obvious (to me).

packager/mux/mux.go
```diff
@@ -37,82 +37,134 @@ import (
...
+			{"/healthz", expectNoSuffix, healthz, "healthz"},
+			{"", return404, nil, "handler_not_assigned"},
+		}}
 }
 
 func tryTrimPrefix(s, prefix string) (string, bool) {
 	sLen := len(s)
 	trimmed := strings.TrimPrefix(s, prefix)
-	return trimmed, len(trimmed) != sLen
+	return trimmed, len(prefix)+len(trimmed) == sLen
```
twifkak, line 116: Do you have an example (or test) to demonstrate the need for this change? I'm not seeing it; AFAICT [TrimPrefix](https://golang.org/pkg/strings/#TrimPrefix) only returns the original string or the correctly-trimmed string.

Also, the previous version seemed clearer to me, but I could be convinced otherwise.

packager/mux/mux.go
```diff
@@ -37,82 +37,134 @@ import (
...
+			{"/priv/doc/", expectSignerQuery, signer, "signer"},
+			{"/priv/doc", expectNoSuffix, signer, "signer"},
+			{"/amppkg/cert/", expectCertQuery, certCache, "certCache"},
+			{"/amppkg/validity", expectNoSuffix, validityMap, "validityMap"},
+			{"/healthz", expectNoSuffix, healthz, "healthz"},
+			{"", return404, nil, "handler_not_assigned"},
+		}}
 }
 
 func tryTrimPrefix(s, prefix string) (string, bool) {
```
twifkak, line 113: While you're here, add a comment to describe the return values. :)

packager/mux/mux.go
```diff
@@ -37,82 +37,134 @@ import (
...
-			params["signURL"] = suffix[1:]
-			if req.URL.RawQuery != "" {
-				params["signURL"] += "?" + req.URL.RawQuery
-			}
-			this.signer.ServeHTTP(resp, req)
-		} else {
-			http.NotFound(resp, req)
+
+	// Find the first matching routing rule.
+	matchingRule := (*routingRule)(nil)
```
twifkak, line 125: Since you're specifying the type anyway, I would write this as:
```
var matchingRule *routingRule
```
instead of as a cast. (Go default-initializes to the zero value, which for pointers is nil.)

packager/mux/mux.go
```diff
@@ -37,82 +37,134 @@ import (
...
+}
+
+// expectCertQuery is a URL Path Suffix Validator specific to cert requests.
+func expectCertQuery(suffix string, req *http.Request, params *map[string]string, errorMsg *string, errorCode *int) {
+	unescaped, err := url.PathUnescape(suffix)
+	if err != nil {
+		*errorMsg, *errorCode = "400 bad request - bad URL encoding", http.StatusBadRequest
+	} else {
+		(*params)["certName"] = unescaped
+	}
```
twifkak, line 93: Thanks for keeping the logic of these so close to the original, so the refactoring is easy to follow!

packager/mux/mux.go
```diff
@@ -37,82 +37,134 @@ import (
...
+	// Prepare the handler.
+	var handlerFunc http.Handler
+	if errorCode != 0 {
+		handlerFunc = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { http.Error(w, errorMsg, errorCode) })
 	} else {
-		http.NotFound(resp, req)
+		handlerFunc = matchingRule.handler
 	}
+
+	handlerFunc.ServeHTTP(resp, req)
```
twifkak, line 159: Optionally, write this as:

```
if errorCode == 0 {
  matchingRule.handler.ServeHTTP(resp, req)
} else {
  http.Error(w, errorMsg, errorCode)
}
```

Benefits (IMO):
 1. Gets rid of a double-negative (the `else` of a `!=`).
 2. Inlines an abstraction that's only used once (`handlerFunc`).
 3. More concise.

MichaelRybak:
@twifkak Devin, thanks for the review! I've addressed your comments. Would you mind having another look?

packager/mux/mux.go
twifkak: I see you met in the middle (did 1 but not 2). That's fine by me, but I'm curious why (I'm betting you thought about it more than me, and I like learning things).

PR #422, opened 2020-05-12
Add end-to-end latency Prometheus metric.

MichaelRybak:


packager/mux/mux.go
```diff
@@ -121,14 +121,29 @@ func New(certCache http.Handler, signer http.Handler, validityMap http.Handler,
 	}
 }
 
-var promTotalRequests = promauto.NewCounterVec(
+// Prometheus counter that observes total requests count.
```
banaag, line 124: Comment should be "promRequestsTotal is a promotheus counter ..."
https://golang.org/doc/effective_go.html#commentary

packager/mux/mux.go
```diff
@@ -121,14 +121,29 @@ func New(certCache http.Handler, signer http.Handler, validityMap http.Handler,
...
+var promRequestsTotal = promauto.NewCounterVec(
 	prometheus.CounterOpts{
 		Name: "total_requests_by_code_and_url",
 		Help: "Total number of requests by HTTP code and URL.",
 	},
 	[]string{"code", "handler"},
 )
 
+// Prometheus summary that observes requests latencies.
```
banaag, line 133: ditto about comment

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/ampproject/amppackager
 go 1.13
 
 require (
+	bou.ke/monkey v1.0.2
```
twifkak, line 6: We cannot use this library: https://github.com/bouk/monkey/blob/master/LICENSE.md

MichaelRybak: Hm, here's a derived library with a MIT liense: https://github.com/undefinedlabs/go-mpatch

I'm hesitant about this anyway. Unit testing of metrics seems unnecessary in general, especially with the hacks that introduce unintended usage of the language. WDYT?

MichaelRybak: Also, here's a quote from an online discussion: Replacing one function with another in this manner will cause memory corruption because the map the garbage collector uses to identify local variables of a function will no longer match.

https://www.reddit.com/r/golang/comments/30try1/monkey_patching_in_go/
https://news.ycombinator.com/item?id=22442170

MichaelRybak: Probably I'd propose to not test it at all. The second option would be to not patch the Since function, and re-implement the comparison of result to make it not strict (because the latencies  won't match perfectly).

MichaelRybak: Given this a bit more thought, I think it still might be worth smoke-testing the latencies measurement after all. I'd suggest I drop the time patching, and test the calculated metrics in some way (so basically my option 2 above). I'll go ahead and implement this, still open to suggestions/input.

MichaelRybak: @twifkak @banaag Per my previous comment I've dropped time patching and am now testing the latencies metric output, with the exception of the exact latency value (because it's non deterministic). I do test that it's non-zero though.

twifkak: Option 2 SGTM.

An alternative could be to implement a variant of [InstrumentHandlerDuration](https://github.com/prometheus/client_golang/blob/ef4f0376f55260aebd309b89b91c641f330b061c/prometheus/promhttp/instrument_server.go#L61) that can be stubbed out for test. Simple version could be something like:

```go
package mux

func InstrumentHandlerDuration(obs, next) {
  if inTest {
    return http.HandlerFunc(func(w, r) {
      next.ServeHTTP(w, r)
      obs.With(something_with_labels_I_dunno).Observe(5)
    }
  } else {
    return promhttp.InstrumentHandlerDuration(obs, next)
  }
}
```

or even simpler, replace the `obs.With...` stuff with like `timesCalled++` on a global var.

(A bug report to Prometheus may be in order; this seems like useful feedback to them.)

MichaelRybak: Thanks!

The "something_with_labels" part depends on the outcome of serving. Specifically to set the right value for "code" label, InstrumentHandlerDuration needs the returned code, and creates a delegate that intercepts the code when written to ResponseWriter. So we'd pretty much need to copy-paste the implementation of InstrumentHandlerDuration. As we would with the the fake clock.

 I've pushed option 2.

MichaelRybak: > A bug report to Prometheus may be in order; this seems like useful feedback to them.

Nice idea. I'll do this after pulling in the PR and will refer to the PR in the report.

PR #425, opened 2020-05-15
Add total gateway requests Prometheus metric.

MichaelRybak:
Add "total_gateway_requests_by_code" metric. It counts total number of amppackager's underlying requests to AMP document server. It breaks the requests down by HTTP response status code.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -331,10 +331,12 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 
 	// According to the check above httpErr is nil, i.e. the gateway request did
 	// succeed. Let Prometheus observe the gateway request along with the response code.
-	// Note that if httpErr was not nil, ServeHTTP wouldn't have reached the code
-	// below. Instead it would let mux's promRequestsTotal observe the non-gateway
-	// request (along with the response code) - by calling LogAndRespond with resp
-	// that mux have decorated with a promRequestsTotal InstrumentHandler.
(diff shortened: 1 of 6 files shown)


packager/signer/signer.go
```diff
@@ -318,6 +329,14 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		}
 	}()
 
+	// According to the check above httpErr is nil, i.e. the gateway request did
```
banaag, line 332: Sorry, I maybe confused.  Where is httpErr referenced above?

packager/signer/signer.go
```diff
@@ -318,6 +329,14 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+	// succeed. Let Prometheus observe the gateway request along with the response code.
+	// Note that if httpErr was not nil, ServeHTTP wouldn't have reached the code
+	// below. Instead it would let mux's promRequestsTotal observe the non-gateway
+	// request (along with the response code) - by calling LogAndRespond with resp
+	// that mux have decorated with a promRequestsTotal InstrumentHandler.
```
twifkak, line 337: Maybe mention specifically that this will be http.StatusBadGateway (502), which gives the reader something to Ctrl-F for in this file, and the key to look for in the promRequestsTotal metric.

MichaelRybak: Done. It can also be 500, so I used 502 as a probable example.

PR #431, opened 2020-06-01
Add gateway response size Prometheus metric.

MichaelRybak:
Another metric per request #357.


twifkak:
actually hold off

packager/signer/signer.go
```diff
@@ -473,6 +487,8 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 		return
 	}
 
+	promGatewayResponseSize.WithLabelValues().Observe(float64(len(fetchBody)))
```
twifkak, line 490: This is the body length after being capped by maxBodyLength above.

 - We should make `>=maxBodyLength` is its own prometheus event so people can know exactly how many hit the limit.
 - (Optional) Also log the value of the `content-length` header, if present? Allow us insight into sizes above maxBodyLength. Should this be a separate SummaryVec? Out of caution, I would lean towards yes; from a quick glance at https://tools.ietf.org/html/rfc7230#section-3.3.2 I can't tell how UAs treat content-length if it differs from actual body length.

PR #437, opened 2020-06-08
Add -staging command line flag.

### LINKED ISSUE #188: Add -staging flag to host cert-url elsewhere

banaag:
FIXIT: Adds explicit -staging command line flag to use or development-mode cert url host. Fixes #188.

### DIFF
--- a/cmd/amppkg/main.go
+++ b/cmd/amppkg/main.go
@@ -43,6 +43,7 @@ import (
 var flagConfig = flag.String("config", "amppkg.toml", "Path to the config toml file.")
 var flagDevelopment = flag.Bool("development", false, "True if this is a development server.")
 var flagInvalidCert = flag.Bool("invalidcert", false, "True if invalid certificate intentionally used in production.")
+var flagStaging = flag.String("staging", "", "URL that overrides the base URL used to host certs, used for testing. Can only be used with -development flag.")
 
 // IMPORTANT: do not turn on this flag for now, it's still under development.
 var flagAutoRenewCert = flag.Bool("autorenewcert", false, "True if amppackager is to attempt cert auto-renewal.")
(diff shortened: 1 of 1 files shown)


cmd/amppkg/main.go
```diff
@@ -43,6 +43,7 @@ import (
 var flagConfig = flag.String("config", "amppkg.toml", "Path to the config toml file.")
 var flagDevelopment = flag.Bool("development", false, "True if this is a development server.")
 var flagInvalidCert = flag.Bool("invalidcert", false, "True if invalid certificate intentionally used in production.")
+var flagStaging = flag.String("staging", "", "URL that overrides the base URL used to host certs, used for testing. Can only be used with -development flag.")
```
twifkak, line 46: I don't think we should limit it to `-development`, since that runs a TLS server. IIRC, @ithinkihaveacat's use-case was to run it in a prod-like setting (behind a rev-proxy).

banaag: Changed, PTAL.

cmd/amppkg/main.go
```diff
@@ -120,8 +121,15 @@ func main() {
...
-	if *flagDevelopment {
-		overrideBaseURL, err = url.Parse(fmt.Sprintf("https://localhost:%d/", config.Port))
+	if *flagStaging == "" {
+		if *flagDevelopment {
+			overrideBaseURL, err = url.Parse(fmt.Sprintf("https://localhost:%d/", config.Port))
+			if err != nil {
+				die(errors.Wrap(err, "parsing development base URL"))
+			}
+		}
+	} else {
```
twifkak, line 131: I think this would read a bit more easily as:

```
if *flagStaging != "" {
  ...
} else if *flagDevelopment {
  ...
}
```

banaag: Good suggestion, done.

PR #438, opened 2020-06-08
FIXIT: Generate a better error for incorrectly generated private key

### LINKED ISSUE #354: Improve error message for non-EC-P256 key
As a small, concrete part of #131. Current message is:

```
failed to parse private key file
github.com/ampproject/amppackager/packager/util.ParsePrivateKey
	github.com/ampproject/amppackager/packager/util/util.go:68
main.main
	github.com/ampproject/amppackager/cmd/amppkg/main.go:101
runtime.main
```

Should suggest using the right key type, and how to do so.

banaag:
Fixes #354

### DIFF
--- a/packager/util/util.go
+++ b/packager/util/util.go
@@ -52,7 +52,8 @@ func ParsePrivateKey(keyPem []byte) (crypto.PrivateKey, error) {
 		var pemBlock *pem.Block
 		pemBlock, keyPem = pem.Decode(keyPem)
 		if pemBlock == nil {
-			return nil, errors.New("invalid PEM block in private key file")
+			return nil, errors.New("invalid PEM block in private key file, make sure to use the right key type. See: https://github.com/WICG/webpackage/tree/master/go/signedexchange#creating-our-first-signed-exchange")
+
 		}
 
 		var err error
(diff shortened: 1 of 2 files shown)


packager/util/util.go
```diff
@@ -52,7 +52,7 @@ func ParsePrivateKey(keyPem []byte) (crypto.PrivateKey, error) {
 		var pemBlock *pem.Block
 		pemBlock, keyPem = pem.Decode(keyPem)
 		if pemBlock == nil {
-			return nil, errors.New("invalid PEM block in private key file")
+			return nil, errors.New("invalid PEM block in private key file, make sure to use the right key type. Generate the key using: openssl ecparam -out priv.key -name prime256v1 -genkey")
```
twifkak, line 55: Maybe instead of inlining the openssl command, link to "step 2 of https://github.com/WICG/webpackage/tree/master/go/signedexchange#creating-our-first-signed-exchange", which includes the necessary command to regenerate the cert, too.

Alternatively, we could update our README or our wiki to include this information, and link to that.

banaag: Done

PR #441, opened 2020-06-17
Click-to-deploy related updates to docker images for AMP Packager

banaag:
Update the docker images of consumer and renewer to have C2D_RELEASE tags.  Add dockerfile for init container for AMP Packager.  Modify base image of renewer and consumer to use alpine instead of scratch.

### DIFF
--- a/deploy/gcloud/Dockerfile.consumer
+++ b/deploy/gcloud/Dockerfile.consumer
@@ -46,8 +46,10 @@ FROM alpine:latest as certs
 
 RUN apk --update add ca-certificates
 
-# Build a small executable from docker scratch.
-FROM scratch
+# Build a small executable from docker alpine. Docker alpine is needed because
+# bash is required to be present by the deployer as it prints the env variables
+# listed below for verifiying the docker image.
+FROM alpine:latest 
 
 # Copy the certs from certs image.
 ENV PATH=/bin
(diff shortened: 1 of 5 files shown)


deploy/gcloud/Dockerfile.consumer
```diff
@@ -62,6 +62,10 @@ EXPOSE 8080
 
 ENV PATH=$PATH:.
 
+# This env var is used by click-to-deploy deployer, right now needs to be
+# manually synced with C2D_RELEASE in the other Docker images for AMP Packager
```
twifkak, line 66: Link to the click-to-deploy repo from somewhere (maybe here).

deploy/gcloud/deploy.sh
```diff
@@ -0,0 +1,107 @@
+# This shell script is used by the docker init container used in the
```
twifkak, line 1: This file seems to have a lot in common with gcloud_up.sh. Is it possible to extract the commonalities into a shared .sh?

deploy/gcloud/Dockerfile.consumer
```diff
@@ -46,8 +46,8 @@ FROM alpine:latest as certs
 
 RUN apk --update add ca-certificates
 
-# Build a small executable from docker scratch.
-FROM scratch
+# Build a small executable from docker alpine.
```
twifkak, line 49: Explain why alpine is necessary here instead of scratch.

deploy/gcloud/Dockerfile.consumer
banaag: Done. Also added in the renewer.

deploy/gcloud/deploy.sh
banaag: Yes, the idea is to eventually remove the similar calls in gcloud_up.sh and use helm combined with the init container to do the container initialization. It's on my todo list but leaving the gcloud* scripts alone for now until click-to-deploy works end-to-end.

deploy/gcloud/Dockerfile.consumer
banaag: Done

deploy/gcloud/deploy.sh
twifkak: Ah, ok, SG.

PR #445, opened 2020-06-25
Refactor serveSignedExchange method.

MichaelRybak:
The first few commands of serveSignedExchange method consume the fetchResp body and are somewhat tricky. The method first consumes all characters up to maxBodyLength limit. If the limit is hit, it proxies full body and exits. If the limit is not hit, it means that the body has been consumed fully and can be signed.

The subsequent calls to `proxy` rely on the background above. Specifically to call `proxy` correctly an engineer needs to remember that fetchResp.Body has already been consumed, and therefore it cannot be re-read. An engineer maintaining the serveSignedExchange method would need that knowledge throughout the whole method, which is somewhat long. There's even a comment to highlight that:

// After this, fetchResp.Body is consumed, and attempts to read or proxy it will result in an empty body.

The new serveSignedExchange_ConsumeCapCompliantFetchResp method encapsulates the logic above: it makes sure the fetch body was read, and takes care of potential io errors and potential hitting of maxBodyLength limit. serveSignedExchange_ConsumeCapCompliantFetchResp returns a `success` bool to notify serveSignedExchange if it's ok to continue signing, which, given with method name, seems straightforward, and makes understanding serveSignedExchange method easier.

Another, similar benefit: the whole capping notion gets hidden into the extracted method, and such structure emphasizes that the part of serveSignedExchange that follows the call to the extracted method works on a non-capped, full body, which is also below the expected limit. It also emphasizes that the body has been fully consumed, so `proxy` doesn't need to read anything else to proxy the full body.

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -80,14 +80,6 @@ var statusNotModifiedHeaders = map[string]bool{
 	"Vary":             true,
 }
 
-// MICE requires the sender process its payload in reverse order
-// (https://tools.ietf.org/html/draft-thomson-http-mice-03#section-2.1).
-// In an HTTP reverse proxy, this could be done using range requests, but would
-// be inefficient. Therefore, the signer requires the whole payload in memory.
-// To prevent DoS, a memory limit is set. This limit is mostly arbitrary,
-// though there's no benefit to having a limit greater than that of AMP Caches.
-const maxBodyLength = 4 * 1 << 20
-
 // The current maximum is defined at:
 // https://cs.chromium.org/chromium/src/content/browser/loader/merkle_integrity_source_stream.cc?l=18&rcl=591949795043a818e50aba8a539094c321a4220c
 // The maximum is cheapest in terms of network usage, and probably CPU on both
@@ -369,7 +361,7 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 
 	if err := this.shouldPackage(); err != nil {
 		log.Println("Not packaging because server is unhealthy; see above log statements.", err)
-		proxy(resp, fetchResp, nil)
+		proxyUnconsumed(resp, fetchResp)
 		return
 	}
 	var act string
@@ -379,7 +371,7 @@ func (this *Signer) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		act, transformVersion = amp_cache_transform.ShouldSendSXG(header_value)
 		if act == "" {
 			log.Println("Not packaging because AMP-Cache-Transform request header is invalid:", header_value)
-			proxy(resp, fetchResp, nil)
+			proxyUnconsumed(resp, fetchResp)
 			return
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -464,15 +464,22 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 	return strings.Join(values, ","), nil
 }
 
-// serveSignedExchange does the actual work of transforming, packaging and signed and writing to the response.
-func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *http.Response, signURL *url.URL, act string, transformVersion int64) {
+// serveSignedExchange_ConsumeCapCompliantFetchResp method consumes the gateway
+// response body, and checks if it's shorter than maxBodyLength and is therefore
+// OK to sign. It does so keeping memory consumption below maxBodyLength. It
+// returns ok iff succeeded and handles the errors if any. Specifically:
```
twifkak, line 470: I'd reword this sentence a bit. Something like:
"If successful, it returns true and stores the result in fetchBody. Otherwise, it returns false and writes a response to resp; the caller should not write to resp in this case."

packager/signer/signer.go
```diff
@@ -464,15 +464,22 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
...
+//   * If failed to read the body, it propagates the error.
+//   * If hit the cap limit, it proxies the content unsigned.
+// In both problematic cases it returns false notifying serveSignedExchange to
+// cancel signing.
+func serveSignedExchange_ConsumeCapCompliantFetchResp(fetchBody *[]byte, resp http.ResponseWriter, fetchResp *http.Response) (success bool) {
```
twifkak, line 475: I'm not wild about this function name, but I don't have a better alternative. :shrug: Probably it's because the functionality is a bit hard to describe (three different cases). But I don't have a better alternative how to divide responsibilities between functions, either. :shrug:

MichaelRybak: I felt the same. I think I now have a better idea, please have a look.

packager/signer/signer.go
```diff
@@ -614,18 +626,39 @@ func (this *Signer) serveSignedExchange(resp http.ResponseWriter, fetchResp *htt
 	}
 }
 
-// Proxy the content unsigned. If body is non-nil, it is used in place of fetchResp.Body.
-// TODO(twifkak): Take a look at the source code to httputil.ReverseProxy and
-// see what else needs to be implemented.
-func proxy(resp http.ResponseWriter, fetchResp *http.Response, body []byte) {
-	for k, v := range fetchResp.Header {
+func proxyUnconsumed(resp http.ResponseWriter, fetchResp *http.Response) {
```
twifkak, line 629: I like this; reads a lot better.

packager/signer/signer.go
```diff
@@ -464,46 +456,64 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
...
+type SXGParams struct {
+	signURL          *url.URL
+	act              string
+	transformVersion int64
+}
+
+// consumedFetchResp stores the fetch response in memory - including the
+// consumed body, not a stream reader. MICE requires the sender to process its
+// payload in reverse order
+// (https://tools.ietf.org/html/draft-thomson-http-mice-03#section-2.1). In an
```
twifkak, line 468: Mention the fact that `transformer.Process` requires the whole contents in memory, too.

packager/signer/signer.go
```diff
@@ -464,46 +456,64 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
...
+
+// maxSignableBodyLength is the signable payload length limit. If not hit, the
+// signer will load the payload into consumedFetchResp and sign it. If hit, the
+// signer won't sign the payload, but will proxy it in full by streaming it.
+// This way the signer limits per-request memory usage, making amppackager more
+// predictable provisioning-wise. The limit is mostly arbitrary, though there's
+// no benefit to having a limit greater than that of AMP Caches.
+const maxSignableBodyLength = 4 * 1 << 20
+
+func (this *Signer) consumeAndSign(resp http.ResponseWriter, fetchResp *http.Response, params *SXGParams) {
```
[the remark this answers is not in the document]
banaag: consumeAndSignIfMax? (sorry I'm making it worse)

PR #453, opened 2020-07-21
Sync from Google

honeybadgerdontcare:
This is an automatically generated pull request.


honeybadgerdontcare:
@twifkak perhaps this isn't ready yet ;D feel free to close.

twifkak:
Oh, fun. Error caused by https://github.com/stretchr/testify/commit/95a9d909e98735cd8211dfc5cbbb6b8b0b665915. Reported in https://github.com/stretchr/testify/issues/981. Possibly won't be fixed because of https://github.com/stretchr/testify/commit/6a6c303c3cf499bdb3c3ab72464dc241416e5f80.

Getting Travis to fetch the version we ask for in https://github.com/ampproject/amppackager/blob/bf061bede4b0280654c25b19fc3cc34e25d13cf0/go.mod#L25

seems to be a difficult maneuver, because different versions of `go get` behave differently, and some versions error out when you pass the flags that the others need. There might be a way to get Travis to pass version-specific build flags. I'm trying.

honeybadgerdontcare:
// cc @Gregable as fyi due to rotation

twifkak:
OK, I guess this would need to be rebased to pick up the .travis.yml changes. Easiest thing is probably to close it and let @Gregable generate a new one as part of his rotation.

PR #458, opened 2020-07-31
Update vendor files.

MichaelRybak:


MichaelRybak:
@banaag Thanks. I just ran "go mod tidy" followed by "go mod vendor", and it didn't update any files.

Issue #314, opened 2019-05-30
Add a GitHub PR check for transformers/


twifkak:
Ensure that no PR includes a change to `transformers/` unless it is a sync from Google (e.g. #309).

(Ideally, we'll make GitHub the source of truth for `transformers/` eventually, but that's a harder project.)

alin04:
Changes in github to transformers is supported. But the author information may be lost (I'd have to double check the configuration on the sync script).

twifkak:
Ah, yes, the PR check should block merging the PR, not mailing it out. The wacky [PR-to-Google](https://github.com/google/copybara/blob/master/docs/reference.md#git.github_pr_origin), [Google-to-GitHub](https://github.com/google/copybara/blob/master/docs/reference.md#git.github_pr_destination) flow should be supported, still.

twifkak:
A small amount of research to be done here, but should be pretty quick. A couple of examples to follow:
 - https://github.com/qawolf/actions-test/blob/master/.github/actions/create-status-check/index.js, per https://github.community/t/github-actions-as-status-checks/16666/5
 - https://github.com/mitchheddles/github-action-status-check, per https://www.google.com/search?q=github+actions+status+checks

Alternatively, I think we could host our own server on the free tier of some cloud? See https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks and https://developer.github.com/v3/guides/building-a-ci-server/. But this seems like unnecessary complexity.

PR #462, opened 2020-08-12
HtmlFormat sync with validator.proto

honeybadgerdontcare:
These two files were meant to be kept in sync but it appears the amppackager request.proto was missing ACTIONS. ACTIONS is now deprecated so adding in comment that 5 is reserved.

github.com/ampproject/amppackager/transformer/request/request.proto
github.com/ampproject/amphtml/validator/validator.proto

### DIFF
--- a/transformer/request/request.proto
+++ b/transformer/request/request.proto
@@ -36,6 +36,8 @@ message Request {
 
   // This should be kept in sync with HtmlFormat.Code in
   // github.com/ampproject/amphtml/validator/validator.proto.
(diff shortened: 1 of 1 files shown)


twifkak:
(I just realized this is undocumented. Wrote up https://github.com/ampproject/amppackager/wiki/Contributing#working-on-transformers. We also have #314 to check automatically.)

honeybadgerdontcare:
Closing in favor of cl/326552765

PR #475, opened 2020-10-13
Add arbitrary preload attributes to Link header

### LINKED ISSUE #461: Is preloadimage working as intended?
I noticed that `Context.Preloads` is populated in https://github.com/ampproject/amppackager/blob/7d23ba92858535987446c874440dc123e85f557c/transformer/transformers/preloadimage.go#L55 and https://github.com/ampproject/amppackager/blob/7d23ba92858535987446c874440dc123e85f557c/transformer/transformers/preloadimage.go#L129, but AFAICT it is never read from.
 
Instead, https://github.com/ampproject/amppackager/blob/7d23ba92858535987446c874440dc123e85f557c/transformer/transformer.go#L320 computes a list of preloads based only on `<script src>` and `<link rel=stylesheet href>` elements. (This is in turn consumed by https://github.com/ampproject/amppackager/blob/7d23ba92858535987446c874440dc123e85f557c/packager/signer/signer.go#L541.)

Do we need to finish integrating `Context.Preloads` into the link header? Or am I missing the part that does this?

jridgewell:
Depends on cl/336023916 being sync'd first.

Closes #461 

/cc @twifkak

### DIFF
--- a/packager/signer/signer.go
+++ b/packager/signer/signer.go
@@ -447,7 +447,7 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 			return "", errors.Errorf("Missing `as` attribute for preload URL: %q\n", preload.Url)
 		}
 
-    valid := true
+		valid := true
 		var value strings.Builder
 		value.WriteByte('<')
 		value.WriteString(u.String())
@@ -457,16 +457,16 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 			value.WriteByte(';')
 			value.WriteString(attr.Key)
 			value.WriteByte('=')
-      qoutedVal, err := util.QuotedString(attr.Val);
-      if err != nil {
-        valid = false
-        break;
-      }
-			value.WriteString(qoutedVal)
+			quotedVal, err := util.QuotedString(attr.Val)
+			if err != nil {
+				valid = false
+				break
+			}
+			value.WriteString(quotedVal)
 		}
-    if valid {
-      values = append(values, value.String())
-    }
+		if valid {
+			values = append(values, value.String())
+		}
 	}
 	return strings.Join(values, ","), nil
 }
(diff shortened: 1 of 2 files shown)


packager/signer/signer.go
```diff
@@ -452,6 +452,13 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
 		value.WriteString(u.String())
 		value.WriteString(">;rel=preload;as=")
 		value.WriteString(preload.As)
+		for _, attr := range preload.GetAttributes() {
+			value.WriteByte(';')
+			value.WriteString(attr.Key)
+			value.WriteString(`="`)
+			value.WriteString(strings.ReplaceAll(attr.Val, `"`, `\"`))
```
twifkak, line 459: Use util.QuotedString instead: https://github.com/ampproject/amppackager/blob/ca6f35a1b2e65f02f8e55c9e9bc218f25b801dc8/packager/util/http.go#L102

(And I guess just skip the whole href if any of its param values are unquotable, for defense-in-depth against bugs in preloadimage.go.)

twifkak (changes requested):

packager/signer/signer.go
jridgewell: Done.

packager/signer/signer.go
```diff
@@ -447,12 +447,26 @@ func formatLinkHeader(preloads []*rpb.Metadata_Preload) (string, error) {
...
 		value.WriteByte('<')
 		value.WriteString(u.String())
 		value.WriteString(">;rel=preload;as=")
 		value.WriteString(preload.As)
-		values = append(values, value.String())
+		for _, attr := range preload.GetAttributes() {
+			value.WriteByte(';')
+			value.WriteString(attr.Key)
+			value.WriteByte('=')
+      qoutedVal, err := util.QuotedString(attr.Val);
```
twifkak, line 460: s/qouted/quoted/

and fix indentation (either manually or by running `gofmt`)

twifkak (changes requested):
cl/337974534 (né cl/336023916) is merged as of #476.

jridgewell (commented):
Rebased.

twifkak (approved):

twifkak:
Thanks! I'll squash-and-merge now.

PR #488, opened 2020-11-20
Fix EAB account workflow to check for existing accounts

banaag:
Fix EAB account workflow to check for existing accounts using ResolveAccountByKey.

### DIFF
--- a/packager/certfetcher/certfetcher.go
+++ b/packager/certfetcher/certfetcher.go
@@ -119,15 +119,18 @@ func New(email string, eabKid string, eabHmac string, certSignRequest *x509.Cert
 		}
 	}
 
-	// Theoretically, this should always be set to false as users should have pre-registered for access
-	// to the ACME CA and agreed to the TOS.
-	// TODO(banaag): revisit this when trying the class out with Digicert CA.
+	var reg *registration.Resource
 	if !shouldRegister {
 		acmeUser.Registration = new(registration.Resource)
+	} else if reg, err = client.Registration.ResolveAccountByKey(); err == nil {
+		// Check if we already have an account.
+		acmeUser.Registration = reg
 	} else {
-		var reg *registration.Resource
-		var err error
-
+		// We need to reset the LEGO client after calling Registration.ResolveAccountByKey().
+		client, err = lego.NewClient(config)
+		if err != nil {
+			return nil, errors.Wrap(err, "Obtaining LEGO client.")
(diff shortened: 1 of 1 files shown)


packager/certfetcher/certfetcher.go
```diff
@@ -119,15 +119,16 @@ func New(email string, eabKid string, eabHmac string, certSignRequest *x509.Cert
 		}
 	}
 
-	// Theoretically, this should always be set to false as users should have pre-registered for access
-	// to the ACME CA and agreed to the TOS.
-	// TODO(banaag): revisit this when trying the class out with Digicert CA.
-	if !shouldRegister {
```
twifkak, line 125: Remove the shouldRegister param; no longer used.

packager/certfetcher/certfetcher.go
```diff
@@ -119,15 +119,16 @@ func New(email string, eabKid string, eabHmac string, certSignRequest *x509.Cert
...
-		acmeUser.Registration = new(registration.Resource)
+	var reg *registration.Resource
+	// Check if we already have an account.
+	if reg, err = client.Registration.ResolveAccountByKey(); err == nil {
+		acmeUser.Registration = reg
 	} else {
-		var reg *registration.Resource
-		var err error
-
+		// We need to reset the LEGO client.
```
twifkak, line 127: Is it feasible to add unit tests for this change? e.g. To prove that it's necessary to reset the client.

twifkak (changes requested):

packager/certfetcher/certfetcher.go
banaag: Unfortunately, no.  I tested it against real CAs and it is necessary.

packager/certfetcher/certfetcher.go
banaag: Actually, we need to keep it so that we can skip registration during tests. Modified the code flow accordingly.

packager/certfetcher/certfetcher.go
twifkak: OK. Add "after calling Registration.ResolveAccountByKey()" to the comment (if that's right).

twifkak (approved):

packager/certfetcher/certfetcher.go
banaag: Done.

PR #491, opened 2020-12-23
Allow the use of AMP LTS releases

pierres:
The AMP Packager overrides the CSP header and only allows AMP modules from the /v0/ path and not /lts/. This breaks AMP pages using the LTS version as the browser blocks the AMP scripts:

Refused to load the script 'https://cdn.ampproject.org/lts/v0.js' because it violates the following Content Security Policy directive: "script-src blob: https://cdn.ampproject.org/rtv/ https://cdn.ampproject.org/v0.js https://cdn.ampproject.org/v0/ https://cdn.ampproject.org/viewer/". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.


twifkak:
Sorry for the delay! This change looks good, but before merging, we'll want to update the Google AMP SXG cache [requirements](https://github.com/ampproject/amppackager/blob/releases/docs/cache_requirements.md) to allow either the old value or the new value. I filed http://b/177370506 to track; will update here once that is deployed.

twifkak:
The fix to http://b/177370506 landed in prod. I tested that a server patched with this PR successfully makes its way through the cache. I'll approve and merge this now. Thanks for this contribution!

twifkak (approved):

twifkak:
Merged into `master`. (You'll need to use `git clone -b` instead of `go get` until our next release.)

PR #498, opened 2021-02-23
remove duplicate test constant

honeybadgerdontcare:
Removes LinkStylesheetGoogleFont in favor of the exactly same valued LinkGoogleFont


banaag (approved):

twifkak (changes requested):
This change should happen inside Google ("source of truth" for transformer code) and then get sync'd out.

PR #515, opened 2021-03-26
Validate OCSP response ProducedAt.

### LINKED ISSUE #514: Validate OCSP ProducedAt
When receiving an OCSP response, AMP Packager should validate that its ProducedAt is within the NotBefore/NotAfter of the cert.  I think the place to do that is [here](https://github.com/ampproject/amppackager/blob/f36311a1de3c0437debf0f6d1c3a0bf6666c7a0a/packager/certcache/certcache.go#L636-L659). (It looks like the `ocsp.ParseResponseForCert` function doesn't do this already.)

twifkak:
Don't accept an OCSP response if its ProducedAt is outside the leaf cert's
NotBefore/NotAfter validity window.

Fixes #514.


twifkak:
To make it easier to review, you may want to look at [this commit](https://github.com/ampproject/amppackager/pull/515/commits/9212c4e59142a9b9080986a7a3727835b753ff5a) separately from the vendor update.

packager/certcache/certcache.go
```diff
@@ -429,9 +429,9 @@ func (this *CertCache) readOCSP(allowRetries bool) ([]byte, time.Time, error) {
 
 // Print # of retries, wait for specified time and returned updated wait time.
 func waitForSpecifiedTime(waitTimeInMinutes int, numRetries int) int {
-	log.Printf("Retrying OCSP server: retry #%d\n", numRetries)
+	log.Printf("Retrying OCSP server: retry #%d", numRetries)
```
banaag, line 432: Did you remove the \n on purpose? Did you mean to change it to Println then?

packager/certcache/certcache_test.go
```diff
@@ -111,8 +112,10 @@ func (this *CertCacheSuite) TearDownSuite() {
 
 func (this *CertCacheSuite) SetupTest() {
 	this.fakeClock = pkgt.NewFakeClock()
+        this.fakeClock.SecondsSince0 = pkgt.B3Certs[0].NotBefore.Sub(time.Unix(0, 0))
+        now := this.fakeClock.Now()
 	var err error
-	this.fakeOCSP, err = FakeOCSPResponse(this.fakeClock.Now())
+	this.fakeOCSP, err = FakeOCSPResponse(now, now.Add(1*time.Minute))
```
banaag, line 118: Could you add a comment here as to why a minute was added?

banaag (approved):

packager/certcache/certcache.go
twifkak: Yeah, the \n is redundant. I saw this done on webpkg. It's a difference between `fmt` and `log`: https://golang.org/pkg/log/#pkg-overview:~:text=Every%20log%20message%20is%20output%20on,newline%2C%20the%20logger%20will%20add%20one.

packager/certcache/certcache_test.go
twifkak: Good idea. Decided to extract it into the FakeOCSPResponse function so there was only one place to document it.

PR #534, opened 2021-05-13
Fix EAB/HMAC flow.  Add instructions on how to get old style ACME Directory URL from DigiCert.

banaag:
Fix EAB/HMAC flow.  The lego client was being reset, losing all it's challenge solvers after ResolveAccountByKey is called."

### DIFF
--- a/packager/certfetcher/certfetcher.go
+++ b/packager/certfetcher/certfetcher.go
@@ -73,13 +73,7 @@ func New(email string, eabKid string, eabHmac string, certSignRequest *x509.Cert
 	config.CADirURL = acmeDiscoURL
 	config.Certificate.KeyType = certcrypto.EC256
 
-	// A client facilitates communication with the CA server.
-	client, err := lego.NewClient(config)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


packager/certfetcher/certfetcher.go
```diff
@@ -159,6 +130,52 @@ func New(email string, eabKid string, eabHmac string, certSignRequest *x509.Cert
 	}, nil
 }
 
+func setupChallenges(client *lego.Client, httpChallengePort int,
+	httpChallengeWebRoot string, tlsChallengePort int,
+	dnsProvider string) (*lego.Client, error) {
```
twifkak, line 135: Either:
1. Don't take a `client`; do the `lego.NewClient` call in here.
2. Just return `error` and add a function docstring that it mutates `client` in-place.

I lean weakly towards 1 since it eliminates a bit more duplication.

twifkak (changes requested):

packager/certfetcher/certfetcher.go
banaag: Done.

twifkak (approved):

Issue #456, opened 2020-07-27
Rename master branch


twifkak:
I feel we should rename it to something else. There seem to be both strong [connotational](https://tools.ietf.org/html/draft-knodel-terminology-02) and [etymological](https://mail.gnome.org/archives/desktop-devel-list/2019-May/msg00066.html) arguments for it.

`main` seems to be the canonical choice nowadays. (I'm open to other options, e.g. `dev` or `trunk`, if there's a good reason.)

[Guidance](https://github.com/github/renaming) is to wait until later, but of the issues listed there, the only affected one is protected branches, AFAIK. Easy enough to change in [settings](https://github.com/ampproject/amppackager/settings/branches). I think the remaining toil is getting everybody to update their local clients and their open PRs and topic branches. Probably best to do this at a time when there's not much action on the repo.

TODO: Find some nice thorough howto. (There seem to be a few out there.)

twifkak:
Should look at ampproject/amphtml#32195 as an example of a successful large-scale migration.

rsimha:
> TODO: Find some nice thorough howto. (There seem to be a few out there.)

[Here's the workflow](https://docs.google.com/document/d/1BN8CkM3b2nydENRI2X1tE6nCxcaeVliUIaC2e-mXZbg) we used for `amphtml` and friends. Some parts might be too heavy-duty for this repo, but figured I'd share in case it's useful.

banaag:
- Renamed master branch to main per https://docs.github.com/en/github/administering-a-repository/managing-branches-in-your-repository/renaming-a-branch
- Renamed fork master branch to main.
- Updated local .git aliases
- Updated local repo:
$ git branch -m master main
$ git fetch origin
$ git branch -u origin/main main

PR #538, opened 2021-05-25
Update docker files to use main instead of master branch.

banaag:

### DIFF
--- a/deploy/gcloud/Dockerfile.consumer
+++ b/deploy/gcloud/Dockerfile.consumer
@@ -27,8 +27,8 @@ RUN apt-get update \
 
 WORKDIR /data
 
-# Run this if you clone from master branch.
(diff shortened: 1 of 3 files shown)


deploy/gcloud/Dockerfile.consumer
```diff
@@ -28,7 +28,7 @@ RUN apt-get update \
 WORKDIR /data
 
 # Run this if you clone from master branch.
```
twifkak, line 30: s/master/main/ here too

twifkak (approved):

twifkak:
Addresses #456.

deploy/gcloud/Dockerfile.consumer
banaag: Done.

twifkak (approved):

PR #542, opened 2021-07-07
Bump Prometheus library

SuperQ:
Update Prometheus client library to the latest release.


SuperQ:
Ugh, more CLAs

MichaelRybak (approved):

MichaelRybak:
Thank you Ben.

twifkak:
@SuperQ Thanks for the fix! Looks like it's blocked on CLA. Is the bot busted or did you not sign? LMK if you'd prefer we submit our own fix.

SuperQ:
I had to find the time outside of work to read through the agreement. It seems reasonable, so I have signed it.

twifkak:
@SuperQ Frustration understood. I just wanted to make sure our bot wasn't stuck (and the offer to take over the PR was genuine). The choice of CLA or DCO is mostly beyond my influence (or even knowledge), but if you're curious, I recall that a discussion took place on https://github.com/ampproject/meta-tsc/issues/25.

Anyway, I'll merge now, thanks again!

SuperQ:
Thanks!

One thing I noticed is that the default branch for this project isn't `main`, should I have made this PR against `releases`? Or do I need to make a separate PR for that? It seems like `main` is a bit out of sync.

twifkak:
`main` is correct; it's the dev branch. No need to do anything else unless it's worth a cherry-pick. `main` gets merged into `releases` every ~2 months after integration testing (with the AMP validator and with Google-internal SXG validation code).

[Side note: It is confusing that `releases` is the default branch; I'd like to change that. This is because, AFAIK, `go get` doesn't support support downloading non-default branches ([source](https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/cmd/go/internal/vcs/vcs.go#L157)). But maybe the new module-aware `go install` does? If so, it may be worth dropping support for older versions of Go.]

SuperQ:
Yes, if you make the release tags semver you can use `go install github.com/ampproject/amppackager@v1.0.0`

You can also use a `latest` string there and it will pick the highest semver.

But I would stay at v0.1.0 or v1.0.0 for the new tags.

twifkak:
Cool, thanks for the tip. Filed #544 to track this.

PR #547, opened 2021-07-13
Enable pprof for debugging

SuperQ:
In order to provide various runtime code profiling featuers, add the
standard Go pprof endpoint. This will allow investigations into
excessive CPU and memory use.


twifkak (approved):
Nice! Thanks. Will merge.

PR #548, opened 2021-07-13
Update various Go modules

SuperQ:
Update several Go modules to current releases. Un-pin some hard-coded
overrides that don't seem to impact build/test.

Add `update-go-deps` make target to simplify updating.


SuperQ:
I tried to update `github.com/WICG/webpackage`, I fixed one struct rename, but it makes the tests panic. Still investigating.

SuperQ:
@twifkak Ok, this one is ready to go too.

Makefile
```diff
@@ -25,3 +25,12 @@ build: amppkg
 .PHONY: amppkg
 amppkg:
 	go build $(GOOPTS) -ldflags "$(VERSION_LDFLAGS)" -o amppkg ./cmd/amppkg/...
+
+.PHONY: update-go-deps
+update-go-deps:
+	@for m in $$(go list -mod=readonly -m -f '{{ if and (not .Indirect) (not .Main)}}{{.Path}}{{end}}' all); do \
```
twifkak, line 31: What's the difference between this for loop and `go get -u` without a package arg? From `go help module-get`:

> In particular, 'go get -u' and 'go get -u=patch' update all the dependencies of that package.

(If the answer is non-obvious, then please add a comment in the Makefile.)

twifkak (commented):
Thanks! One possibly naive question:

Makefile
SuperQ: This is meant to skip the `// indirect` modules. Since they typically require extra care in updating, or are automatically updated when you get other packages.

twifkak: Ah OK. And I guess in this particular case you reverted the updates it made to WICG/webpackage?

twifkak (approved):
Thanks, LGTM, merging.

SuperQ:
Yes, WICG/webpackage is going to take some more work to update.

PR #555, opened 2021-07-16
Update WICG/webpackage

SuperQ:
Update to latest WICG/webpackage module.

Signed-off-by: SuperQ <superq@gmail.com>

PR #563, opened 2021-07-30
Upgrade WICG/webpackage dependency.

twifkak:
To be compatible with the new API:
- Rename CertChainItem to AugmentedCertificate.
- Remove PEM decoding from util.ParsePrivateKey as that is now handled by
  signedexchange.ParsePrivateKey.
- Add a separate parser for an RSA private key as
  signedexchange.ParsePrivateKey no longer allows that type. Likewise, update
  tests to use ECDSA key as appropriate.

/cc @SuperQ

Obviates #551 and #555.


banaag (approved):

SuperQ:
Nice, thanks!

Lots of stuff is updated now, maybe time for a release?

twifkak:
Good idea; in progress.

PR #603, opened 2021-12-16
Add note to TOML regarding retries during failure case.

banaag:
Add a note for cases when packager keeps retrying to retrieve new cert when there's a scenario that requires user intervention on their CA account.


deploy/gcloud/amppkg_renewer_template.toml
```diff
@@ -1,3 +1,9 @@
+# IMPORTANT NOTE: in the event that cert renewal fails and requires user intervention on the user's CA account,
+# amppackager will continue to request new certificates in the background which will result in multiple orders
+# possibliy queueing up in the user's CA account. Make sure to clear all those orders as you it may result in
```
Gregable, line 3: possibly

Gregable (changes requested):

deploy/gcloud/amppkg_renewer_template.toml
banaag: Done.

Gregable (approved):

PR #693, opened 2023-06-07
Modified from PR 689 to make vet pass.

banaag:
Updated PR 689 to make go vet pass.


packager/mux/mux_test.go
```diff
@@ -234,7 +234,7 @@ func TestParamsIncorrectValueType(t *testing.T) {
 // There is an option that doesn't look safe enough:
 // https://www.reddit.com/r/golang/comments/30try1/monkey_patching_in_go/
 // https://news.ycombinator.com/item?id=22442170.
-func TestPrometheusMetricRequestsLatency(t *testing.T) {
+func DisableTestPrometheusMetricRequestsLatency(t *testing.T) {
```
caoboxiao, line 237: Is this test disabled?

banaag: Yes that's on purpose. There's something wrong with Prometheus upgrade. Will revisit after all the dependencies are addressed.

antiphoton (approved):

